PR #21, opened 2025-04-30
Fix import in schema package docs

### LINKED ISSUE #4: Inconsistent import in code demo for "write a schema package"
Noticed some issues in [the code demo](https://nomad-lab.eu/prod/v1/staging/docs/howto/plugins/schema_packages.html#schemapackage-class) for "Write a schema package"

Some imports are missing (e.g. `MEnum`), and some imports are not used (`ELNAnnotation`/`ELNComponentEnum `)

I would fix this after our docs migration to github

DanielYang59:
Fix import in schema package docs, to fix #4


docs/howto/plugins/schema_packages.md
```diff
@@ -101,6 +100,8 @@ _Subsections_ allow the placement of sections within each other, forming contain
 To use the above-defined schema and create actual data, we have to instantiate the classes:
 
 ```python
+from nomad.metainfo.example import Run
+
 run = Run()
```
DanielYang59, line 105: Not sure if this is the correct import

ladinesa: this will not work as the System def here is different from the one defined in example.py. One should instead define `class Run` here containing the system sub-section

DanielYang59: Thanks for the input Alvin, maybe you could help me clarify this doc a bit? I guess I'm not the right person to write up tutorial for schema package :D

ladinesa: Sure, but I will not be able to work on it in this pr. Perhaps do some changes then i will have a look or put some comments if things are not clear like this one.

DanielYang59: I hope a comment in a9d986e902b9738eb9f5558f4d9a473e4360f2fe is good enough :)

ladinesa: no, this should actually be ```
simulation = Simulation()
system = System()
system.n_atoms = 3
system.atom_labels = ['H', 'H', 'O']
simulation.system.append(system)
```

please also change the output json
simulation.m_to_json(indent=2)
```
{
    "m_def" = "Simulation",
    "systems": [
        {
            "n_atoms" = 3,
            "atom_labels" = [
                "H",
                "H",
                "O"
            ]
        }
    ]
}
```

Comment (DanielYang59) [reply docs/howto/plugins/schema_packages.md:105]: Thanks! in my case the output json seems to be:

```json
{
  "system": [
    {
      "n_atoms": 3,
      "atom_labels": [
        "H",
        "H",
        "O"
      ]
    }
  ]
}
```

ladinesa: ah yes, i think it only outputs `m_def` when you  set `with_root_def` in `m_to_json` which by default is `False` or if the def somehow corresponds to an inheriting section

DanielYang59: Thanks. Is everything look good to you?

DanielYang59: > i think it only outputs m_def when you set with_root_def in m_to_json which by default is False

Ah looks like to have to manually wrap `m_to_dict` with `json_dumps` as [`with_root_def` seems to be a kwarg of `m_to_dict`](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/develop/nomad/metainfo/metainfo.py?ref_type=heads#L1769) which isn't passed to [`m_to_json`](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/develop/nomad/metainfo/metainfo.py?ref_type=heads#L2283-2285)

ladinesa: ah yes you are right.  this was by design i guess as it is cumbersome to distinguish the json and metainfo kwargs.

DanielYang59: Thanks for looking this into (it's so complicated)

ladinesa (approved):

PR #44, opened 2025-06-18
add tutorial page to link workflow tutorial

JFRudzinski:


JFRudzinski:
@ahm531 Thank you for the feedback! I have implemented your suggestions.

In the mean time, I also made some additions to `upload_publish.md` please see what you think. I am starting to work on updating the comp docs and these things I added are general items that I would like to remove from the comp section, but were somewhat missing from the tutorial IMO.

ahm531 (approved):
Very good addition to the upload tutorial page! Thanks. 

Few suggestions and comments below:

##### Typos and corrections: 
- Typo in
`The name of the upload can be **modify** by clicking on the pen icon`
→ Change *modify* into *modified*

- Typo (i guess) in
`API: generates a JSON response **to use** via ....` 
→ Change *to use* into *to be used* 

- Improved phrasing in
`Delete the upload: deletes completely the upload.`
→ Change to: *deletes the upload permanently.*


##### Suggestions for wording: 

- `Manage members: allows users to invite collaborators by **defining** co-authors and reviewers roles.`
I suggest using *assigning* instead of *defining*, since the roles are predefined in the code and this better reflects the user’s action at the GUI (assigning roles).*

- `Sections of the upload overview` 
I recommend to avoid using the word Sections in this context, and reserve its use for the sections of the NOMAD data model. Perhaps we can replace it in the title with *Parts*,  *Segments*, *Components* ....

- `The remainder of the uploads page is divided in **5 sections:**` 
I suggest the following: *The remainder of the uploads page is divided into five segments, each presenting a step in the uploading and publishing process.*

- `displays the files and folder structure **in** the upload. You can add a **README.md in** the root directory and its **content** will be shown above this section.`
I suggest the following modification: 
*displays the files and folder structure of the upload. You can add a README.md file to the root directory and its contents will be shown above this section.*

- ` allows users to make the upload **public before publishing or to share the upload with specific users.**`
I suggest the following modification: 
*allows users to make the upload public or share it with specific users before publishing.*

- `5. Publish: lets **the** user **to** publish the data with or without an embargo (i.e., a waiting period before the data **is** publicly visible).`
I suggest the following: 
*5. Publish: lets users publish data with or without an embargo (i.e., a waiting period before the data are publicly visible).



##### Visual and style alignment: 

- I suggest removing the bullets (the small dots) in the list of the icons and keeping only the icons, as they already function like bullet points.

- To align with the overall style of this page (until we have a writing guide), I suggest to:
        - Capitalize the first word after the colon in lists. 
        - List the items before the colon in Bold. (instead of italic and inline)

PR #47, opened 2025-06-24
Enhance Graph API documentation with introductory and functional details

TLCFEM:
Add introductory documentation for the Graph API, comprehensive details on miscellaneous functionalities, and a section for listing file information. This improves clarity and usability for developers interacting with the API.

### DIFF
--- a/docs/howto/graph-api/misc.md
+++ b/docs/howto/graph-api/misc.md
@@ -1,0 +1,228 @@
+# Miscellaneous Functionalities
+
+Apart from the core functionalities that cover MongoDB database and archive data, there are additional resources that can be fetched.
+In this page, we briefly summarize these functionalities.
+
+## User Information
+
+As all data are associated with the corresponding users, in our mental model, uploads, entries, datasets, etc., are connected to a user node.
+Accordingly, there is a top level special token `users` that can be used to fetch user information.
+Just like uploads/entires introduced previously, a user ID needs to be specified.
+
+For example, to fetch the user information of a user with ID `57aaf068-cdd0-43c1-be51-99e0d425c131`, one can use the following query.
+
+```json hl_lines="3"
+{
+   "users":{
+      "57aaf068-cdd0-43c1-be51-99e0d425c131":{
+         "m_request":{
+            "directive":"plain"
+         }
+      }
+   }
+}
+```
+
+This will return the following response (some fields are omitted).
+
+```json
+{
+   "users":{
+      "57aaf068-cdd0-43c1-be51-99e0d425c131":{
+         "name":"Theodore Chang",
+         "affiliation":"Humboldt-Universität zu Berlin",
+         "user_id":"57aaf068-cdd0-43c1-be51-99e0d425c131",
+         "created":"2022-02-12T22:14:25.151000+00:00",
+         "is_admin":false,
+         "is_oasis_admin":false
+      }
+   }
+}
+```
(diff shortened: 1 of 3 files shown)


docs/howto/graph-api/query.archive.md
```diff
@@ -0,0 +1,706 @@
...
+Each archive thus corresponds to an entry, and the corresponding entry ID can be used as the unique identifier to access the archive.
+In the graph system, the archive is linked to the corresponding entry via the special token `archive`.
+Thus, to access the archive of an entry with ID `example_entry_id`, one can use the following query.
+
+```json hl_lines="4-6"
+{
+   "entries":{
+      "example_entry_id":{
+         "archive":{
+            "m_request":{ "directive":"plain" }
```
lauri-codes, line 20: The `"directive": "plain"`, is a bit mysterious for the reader at this point. Is it the default behaviour? If so, maybe we don't need to include it in these examples? If you think it is better to keep it here, add a small note + link to the `"Resolving References"` section that explains the meaning of this in more detail.

docs/howto/graph-api/misc.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Elasticsearch Metadata
+
+Some of the entry metadata is also stored in `Elasticsearch` indices.
+They are similar to the metadata in the MongoDB database, but with more information.
+To fetch data from the `Elasticsearch` index, one can use the special token `search`.
+The top-level request configuration also supports the `query` field.
+This `query` field shall take a valid `Metadata` query object (which itself also contains a `query` field), see the endpoint `/entries/query` for more details.
+
+The following example lists all entries created in 2025 that are visible to the logged in user.
```
lauri-codes, line 135: The used query means more specifically: "... all entries created after the beginning of year 2025"

docs/howto/graph-api/misc.md
```diff
@@ -0,0 +1,319 @@
...
+      "HcLSOzI89sGMCgYnhhfztI4Z1qxY":{ "entry":{ "entry_create_time":"2025-06-23T16:07:56.364000" } }
+   }
+}
+```
+
+??? note "data duplication"
+    Some fields are both stored in the `MongoDB` database and the `Elasticsearch` index.
+    And it is likely that the desired information can be directly fetched from `Elasticsearch`.
+    In this case, there is no need to navigate to the `MongoDB` database.
+    However, if the request needs to access archive, it has to use `entry -> archive` path if starting from `search`.
```
lauri-codes, line 228: This is an important point. Just for me to know: does the API already skip MongoDB if the requested information is in Elasticsearch?

lauri-codes (commented):
Thanks for extending the docs, looks good. I only had a few small comments. You can resolve them on your own and then merge this.

docs/howto/graph-api/misc.md
TLCFEM: No, it returns a list of "entry metadata", depending on the query and required, it may or may not include metadata stored in mongo.
You can still navigate to the corresponding mongo entry metadata via `entry` token.
But mostly like you can directly get them by assigning `required` field in the search query.

TLCFEM:
I may not have the write access.
Even if I have, the policy requires approving reviews.
So it needs to be approved first, and then if I am granted the write access I will merge it.
Otherwise you can directly merge it.
@lauri-codes

PR #49, opened 2025-07-13
Extending mapping annotation

ndaelman-hu:
This is a first extension on the documentation for the mapping parser, based on interactions with @ladinesa and testing it myself.
I start off on a more **conceptual introduction**, introducing **terminology**, and then pivot to **how-to's**.
I also added illustrations, as well as fix the link to the side bar menu.

TODO: improve integration with technical side by @ladinesa .


docs/howto/customization/mapping_parser.md
```diff
@@ -1,12 +1,130 @@
 # How to write data to archive with MappingParser
 
-`MappingParser` is a generic parser class implemented in
-`nomad.parsing.file_parser/mapping_parser.py` to handle the conversion to and from a
+`MappingParser` is a NOMAD parsing framework for handling tree-like data structures.
+It is extensively used for processing highly structured files, like computational output.
+This framework distributes responsiblities between the file processing (building the source tree) and the mapping into the NOMAD schema.
+Some functionality can be inserted in the mapping, the rest is relegated to *normalization* functionalitites of the schema.
```
ladinesa, line 6: I am not sure if we include the discussion of the infrastructure  here, i suggest you reference the already existing docs on parsing-normalization.

docs/howto/customization/mapping_parser.md
```diff
@@ -1,12 +1,130 @@
...
 this class as 'mapping parser' throughout this section. In the following, the abstract
 properties and methods of the mapping parser are explained. The various implementations of
 the mapping parser are also defined and `Mapper` which is required to convert a
-mapping parser into another mapping parser is explained as well.
+mapping parser into another mapping parser is explained as well. -->
+
+## Fundamentals
+
+The mapping from the source file to the archive is defined by a path (in [JMesPath](https://jmespath.org/) format) to the schema.
+This path is added to `m_annotations`, either overwriting or extending the previous annotations.
```
ladinesa, line 20: you can add reference to the annotations here.

docs/howto/customization/mapping_parser.md
```diff
@@ -1,12 +1,130 @@
...
+Each mapping corresponds to its own dictionary key, and a parser schema may contain multiple in parallel.
```
ladinesa, line 21: Maybe flip this, multiple annotations can be provided with their own key, which can be set prior to mapping.

docs/howto/customization/mapping_parser.md
```diff
@@ -1,12 +1,130 @@
...
+
+In the simplest case, e.g. a _single quantity_, you may simply write:
+
+```python
+<quantity>.m_annotations.setdefault(MAPPING_ANNOTATION_KEY, {}).update(
+    dict(xml=MapperAnnotation(mapper=<jmes_path>))
```
ladinesa, line 27: you may already put multiple annotations here to illustrate your point above.

docs/howto/customization/mapping_parser.md
```diff
@@ -1,12 +1,130 @@
...
+```
+
+### Controlling Repeating Sections
+
+**Subsections** that are defined as **repeating** in the schema are automatically picked up by the mapping parser.
+It will look for any repeating units along the path and instantiate the same number of subsections, i.e. $\text{no. subsections} = \prod_{\text{segments}} \text{repeating path segment}$.
+To compose parallel branches into the same repeating subsection, i.e. $\text{no. subsections} = \sum_{\text{branches}} \text{parallel branch}$:
+
+1. generate different mappings and add them to the annotation `dict`. Each mapping comes with its own unique key.
+1. during the *conversion phase*, select `update_mode="append"`. Manipulate the order in the data via the conversion order of the various maps.
```
ladinesa, line 94: maybe you also include here the various update modes

ladinesa (changes requested):

docs/howto/customization/images/mapping_parser_path.png
EBB2675, line 1: Could we please restyle these diagrams so they match the in-house style we use elsewhere in the NOMAD docs?

JFRudzinski (changes requested):
Thanks for contributing to these docs @ndaelman-hu !

For now I will leave only overarching comments, as I think there are some bigger-picture adjustments to make before going through all the details:

1. This page belongs under the Plugins category. (I will make another PR to help clarify Customization vs. Plugins, and also move the hdf5 page).
2. Please split into a How-to and Explanation. Much of what is here already is the explanation part. The how-to should begin with the basic imports of the tool within a plugin, then can have clear use-case examples for the different parser types (i.e., xml, hdf5, etc) and the different standard usages. It is fine to use the xml for most/all of the examples for now. I can add hdf5 examples later if there is anything unique there.
3. I like the figure style for a talk but I agree that the figures need to be standardized, however, this is an activity which needs to be done for the entire docs (there are plenty of examples of non-standardized figures), and maybe we can assign someone else to do this. Let's work on the rest of the PR first and I will look into this

ndaelman-hu:
> Thanks for contributing to these docs @ndaelman-hu !
> 
> For now I will leave only overarching comments, as I think there are some bigger-picture adjustments to make before going through all the details:
> 
>     1. This page belongs under the Plugins category. (I will make another PR to help clarify Customization vs. Plugins, and also move the hdf5 page).
> 
>     2. Please split into a How-to and Explanation. Much of what is here already is the explanation part. The how-to should begin with the basic imports of the tool within a plugin, then can have clear use-case examples for the different parser types (i.e., xml, hdf5, etc) and the different standard usages. It is fine to use the xml for most/all of the examples for now. I can add hdf5 examples later if there is anything unique there.
> 
>     3. I like the figure style for a talk but I agree that the figures need to be standardized, however, this is an activity which needs to be done for the entire docs (there are plenty of examples of non-standardized figures), and maybe we can assign someone else to do this. Let's work on the rest of the PR first and I will look into this

Agreed with all 3 points. Regarding 1 and 2, do you already have some of this structure applied?

docs/howto/customization/mapping_parser.md
ndaelman-hu: You mean that people provide just the keys, but not the values..? Is that really useful to mention?

PR #108, opened 2025-08-07
Add run-gemini-cli

blueraft:
Adds the default workflow from gemini-cli as described in https://github.com/google-github-actions/run-gemini-cli?tab=readme-ov-file#3-choose-a-workflow


JFRudzinski:
These questions to gemini were placed in the wrong PR, please ignore

PR #118, opened 2025-08-24
Fix and extend api plugin docs

rettigl:
During testing the examples, I noticed several errors and inconsistencies. I also added further documentation on e.g. how to serve static content using this approach.

### DIFF
--- a/docs/howto/plugins/apis.md
+++ b/docs/howto/plugins/apis.md
@@ -37,7 +37,6 @@ Furthermore, it allows you to define a path prefix for your API.
 The entry point should be defined in `*/apis/__init__.py` like this:
 
 ```python
-from pydantic import Field
 from nomad.config.models.plugins import APIEntryPoint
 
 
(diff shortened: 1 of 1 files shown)


docs/howto/plugins/apis.md
```diff
@@ -93,5 +92,39 @@ If you run NOMAD with this plugin following our [Oasis configuration documentati
...
 ```sh
-curl localhost/nomad-oasis/myapi
+curl localhost:8000/nomad-oasis/myapi/
+```
+
+In a nomad-distro-dev environment, the API should be accessible at:
+
+```sh
+curl localhost:8000/fairdi/nomad/latest/myapi/
+```
```
lauri-codes, line 102: Instead of having to account for these two options always in our docs, we should just make it so that both `nomad-distro-template` and `nomad-distro-dev` use the same path by default. I have created an issue here: https://github.com/FAIRmat-NFDI/nomad-distro-dev/issues/75.

lauri-codes (commented):
Thanks a lot for these corrections!

I will do a few minor tweaks and merge this into our docs.

PR #125, opened 2025-09-02
fix actions markdown first level title

DanielYang59:
fix:
```
Error: docs/howto/plugins/actions.md:1 MD041/first-line-heading/first-line-h1 First line in a file should be a top-level heading [Context: "!!! tip "Warning""] https://github.com/DavidAnson/markdownlint/blob/v0.38.0/doc/md041.md
```

### DIFF
--- a/.github/workflows/test-build-and-links.yml
+++ b/.github/workflows/test-build-and-links.yml
@@ -35,13 +35,23 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Check links in markdown files
-        uses: tcort/github-action-markdown-link-check@v1
+      - name: Setup Node.js 20
+        uses: actions/setup-node@v4
         with:
-          use-quiet-mode: 'yes'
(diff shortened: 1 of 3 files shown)


.github/workflows/test-build-and-links.yml
```diff
@@ -35,13 +35,23 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Check links in markdown files
-        uses: tcort/github-action-markdown-link-check@v1
```
DanielYang59, line 39: link check sometimes fail which would fail the pipeline which is quite annoying. However AFAIK the action (and the CLI `markdown-link-check` itself) [on support retry upon 429 too many requests](https://github.com/tcort/github-action-markdown-link-check?tab=readme-ov-file#too-many-requests)

Not sure if there's better way than hard code this retry step?

blueraft (approved):

PR #132, opened 2025-09-17
Remove PR trigger for docs deployment

DanielYang59:
revert https://github.com/FAIRmat-NFDI/nomad-docs/pull/38

This leads to PR changes being directly deployed, is this needed for a reason?

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -4,9 +4,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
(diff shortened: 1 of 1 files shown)


JFRudzinski:
I would like to see what @lauri-codes or @TLCFEM says about this, IMO it can/should be removed, but I wonder what the initial intention was

DanielYang59:
Yes I agree it's better to use a dedicated site to preview the docs, let's merge this and I would work on that in a separate PR?

PR #141, opened 2025-09-23
72 consolidate graph style api pages

### LINKED ISSUE #72: Consolidate How-to Graph-style API into single page
these pages exceed the maximum allowed nested structure for nomad-docs

JFRudzinski:
closes #72 

- consolidates graph API docs into a single page `docs/howto/manage/program/graph_api.py`
- trims down tutorial-like narrative, and removes some unnecessary cross-section idea referencing
- trims sentences and attempts to clarify language
- some stylistic adjustments for consistency with the rest of the docs

@TLCFEM @ndaelman-hu These are very nice docs. My focus was to consolidate the pages to align with the rest of the how-tos, but I tried to improve the clarity along the way. 

For reviewing ease, I duplicated the new text and separated it back into the old individual files. Otherwise you can just launch the docs and read through the new version (single page). I hope this makes sense. Please check that I did not change the meaning of anything. 

Finally, there are a few TODOs left, could you please advise me on how to proceed with those.

### DIFF
--- a/docs/howto/manage/program/publish_python.md
+++ b/docs/howto/manage/program/publish_python.md
@@ -10,13 +10,8 @@
 
 ## Further resources
 
-<<<<<<< HEAD
 - [API Overview](./api.md)
 - [nomad-utility-workflows > How-to Guides > Perform API Calls](https://fairmat-nfdi.github.io/nomad-utility-workflows/how_to/use_api_functions.html){:target="_blank"}
-=======
-- [How-to guides > Programmatic use > Use the API](./api.md)
-- [nomad-utility-workflows > How-to Guides > Perform API Calls](https://fairmat-nfdi.github.io/nomad-utility-workflows/how_to/use_api_functions.html){:target="_blank" rel="noopener"}
->>>>>>> origin/main
 
 ## Uploading, changing metadata, and publishing via python API
 


JFRudzinski:
> > > consolidates graph API docs into a single page
> > 
> > 
> > I was thinking about this but did not pick this approach.
> > Not sure if it is a good idea, because it's going to be too long for a single page.
> 
> I definitely understand the concern, but there has been a consensus to limit the tree depth in the navigation bar. The alternative would be to keep separate pages but without the graph API sub-folder. I think this is not as nice as our priority atm is to help users navigate to the appropriate location.
> 
> Using subsection organization helps somewhat alleviate the long page issue. I think in the future, with a more sophisticated search mechanism and non-linear connections between pages, we could revert to sub-dividing topics into more digestible sections. For now, I believe this is the best approach for a cleaner organization.
> 
> I am open to other suggestions that I have perhaps not considered

@TLCFEM I was thinking about this a bit more. I can see an alternative option that we split the pages as you had before or something similar. But only if we can define a clear context that is not reliant on some overarching super-context, i.e., remove the explicit connection with graph-api in the actual how-to (the graph-api description can be moved to explanation). e.g., "How to query archives", "How to perform fuzzy queries", etc. 

Do you see this as a better approach? The downside is you give less immediate context in terms of connecting the various topics that you discuss there.

TLCFEM:
"How to query archives," "How to perform fuzzy queries," etc., do not stand on their own.
They require prerequisite knowledge of the structure of a graph request.

Those four pages need to be presented in a fixed order, whether combined into a single page or kept separate.
If I had to choose, I would prefer the single-page format.

docs/howto/manage/program/graph-api/query.archive.md
```diff
@@ -605,11 +590,21 @@ As a matter of fact, all references will be recursively resolved such that the r
 }
 ```
 
-### Controlling Reference Resolution
+The first thing to note here is that the reference string has been normalized to `uploads/RzWMitKESo2dQmuE6uQB-Q/entries/x36WdKPMctUOkjXMyV8oQq2zWcSx/archive/run/0/calculation/2`.
+By default, the 'extra' data requested (here, the resolved data) will be added to the response under a fixed path: `uploads/<upload_id>/entries/<entry_id>/archive/<path_to_data>`.
```
ndaelman-hu, line 594: Had to re-read this to understand what "fixed path" meant.


```suggestion
By default, the 'extra' data requested (here, the resolved data) will be added under the response path: `uploads/<upload_id>/entries/<entry_id>/archive/<path_to_data>`. This applies, even when this path differs from the one specified in the original request.
```

docs/howto/manage/program/graph-api/query.archive.md
```diff
@@ -605,11 +590,21 @@ As a matter of fact, all references will be recursively resolved such that the r
...
+This is a valid path that can be used to access the data in the **same** response.
+The motivation is to produce a response that is as self-contained as possible.
+
+The second thing to note is that the target `calculation` contains a further reference `method_ref` that points to the method used for the calculation.
+This reference is also resolved, and the corresponding data is fetched and included in the response.
+In general, whenever the `resolved` directive is specified, all references will be recursively resolved such that the response contains all the data that is reachable from the original reference.
+
+#### Controlling Reference Resolution
 
 However, it is **not always** desired to resolve all references.
```
ndaelman-hu, line 605: I would not use "However" when starting a new sub-chapter.


```suggestion
It is **not always** desired to resolve all references.
```

ndaelman-hu (approved):
Was very interesting to read through. I learned a lot about the graph API.
I mostly pointed out places where I felt some clarification was needed, as well as some typos.

ndaelman-hu:
> Those four pages need to be presented in a fixed order, whether combined into a single page or kept separate. If I had to choose, I would prefer the single-page format.

I'm somewhat ambiguous towards this discussion. I understand limiting the navigation menu's depth, but the current file structure is useful... Considering all sides together, I would consolidate this into a single file, though:

- files have their own navigation bar on the right-hand side.
- graph queries are an advanced, standalone topic that mostly concerns devs. They shouldn't take up too much space in the general exploration menu.

> "How to query archives," "How to perform fuzzy queries," etc., do not stand on their own. They require prerequisite knowledge of the structure of a graph request.

@TLCFEM I agree that context is needed to understand these, but their examples provide quick templates. I would say that your organizational framework more closely matches a "textbook" style.

A few years back, we had an expert at Canonical present his organizational system, [Diataxis](https://diataxis.fr/), that uses a different lay-out. There the context would go under "explainer", and the concrete examples as well as showcases of specialized fields under "how-to".
The rationale behind this approach is allowing for **different tiers of time investment on the reader's side**. Of course, this all depends on how much we continue to use this approach. @JFRudzinski ?

docs/howto/manage/program/graph-api/basics.md
```diff
@@ -1,48 +1,46 @@
...
+    As a result, there are unfortunately no GUI tools available at this time.
 
-## Basic Usage (Data Fetching)
+## Basic Data Fetching
 
-Imagine there is an example upload with the upload ID `example_upload_id`.
-The metadata of this upload is stored in `MongoDB`.
-In our mental model, it is possible to organize all uploads in a tree structure, under the root `uploads`.
-This better helps us to define interactions among different data structures (to be explained later).
+Imagine there is an example upload with the upload ID `<example_upload_id>`. The metadata of this upload is stored in MongoDB.
```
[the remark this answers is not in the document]
TLCFEM: > Still, this statement is technically false.

Why so?

docs/howto/manage/program/graph-api/graph.structure.md
```diff
@@ -64,18 +62,14 @@ Thus, all entries under the upload can be fetched as follows:
...
-    This means it can only be used to represent `upload_id`, `entry_id`, `dataset_id`, etc., for data that follows a fixed schema (`MongoDB`).
-    It won't work in archives, the corresponding metainfo (definitions), and alike.
+    This means it can only be used to represent `upload_id`, `entry_id`, `dataset_id`, etc., for data that follows a fixed schema (e.g., MongoDB).
+    It won't work for archive data, the corresponding metainfo (definitions), and alike.
 
-## Fuzzy Query
+### Fuzzy Querying
 
-The request configuration allows one to perform fuzzy queries to further filter data before fetching, via the `query` and `pagination` fields.
+The request configuration allows one to perform fuzzy queries to further filter data before fetching via the `query` and `pagination` fields:
```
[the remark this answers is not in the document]
ndaelman-hu: I'm asking since "fuzzy fetching" is somewhat defined above, but did not spot the definition for "fuzzy queries".

docs/howto/manage/program/graph-api/graph.drawio
TLCFEM, line 1: Please do not remove this file.
This will be useful for updating the figure.

JFRudzinski: I understand, please see my [discord message](https://discord.com/channels/1201445470485106719/1230108467050713208/1421070558153674832)

In short, I have saved the file to HU Box, as we don't allow unused assets in the repo. Perhaps we need a solution where some templates like this are allowed, I will bring this up, and move it back if that is the case

PR #155, opened 2025-09-25
Add `check_links` to auto insert `{:target="_blank" rel="noopener"}` to external links

### LINKED ISSUE #142: Add `rel="noopener"` to external links
all external links should use: `[](){:target="_blank" rel="noopener"}`

DanielYang59:
close #142


JFRudzinski (approved):
:+1:

.github/workflows/check_markdown_links.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python3
+# /// script
+# requires-python = ">=3.12"
```
blueraft, line 3: This should be part of the `nomad_docs` python package, with `regex` added to the pyproject.toml as a normal dependency.

There can be a test in the tests/ folder that checks if all of the markdown files uses the correct attributes. And a couple of unit tests for the functions added in this file.

blueraft:
The code looks good, can you add a short sentence to the README about this requirement?

PR #170, opened 2025-10-14
Update actions docs for NOMAD 1.4.0

### LINKED ISSUE #126: Improvements in Actions docs
- [ ] Documentation for using actions in dev-distro is missing. Specifically, the need and command for running an action worker for a given queue.

blueraft:
Closes #126 

Pending on changes from https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/tree/encrypt-action-data

### DIFF
--- a/docs/howto/plugins/types/actions.md
+++ b/docs/howto/plugins/types/actions.md
@@ -1,12 +1,5 @@
 # How to define an action
 
-!!! tip "Warning"
-
-    NOMAD Actions are a preview feature available starting from NOMAD version `nomad-lab>=1.3.18.dev89`.
-    Please note that this is an early release: some functionality may change, and you might encounter bugs.
-
-    We’d love your feedback to help us improve NOMAD Actions, please don’t hesitate to reach out and share your thoughts.
-
 Actions allow to define executable workflows in NOMAD. They are an
 alternative to [normalizers](../../../tutorial/custom.md#custom-normalizers) and can
 be configured to use specialized workers instead of the NOMAD internal worker.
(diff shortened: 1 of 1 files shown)


docs/howto/plugins/types/actions.md
```diff
@@ -147,6 +140,9 @@ a collection of activities and workflows that connect them. It also specifies th
 workflows are made available through the `Action`, they can be triggered
 using the `start_action` funtion. This adds a workflow run instance to the specified task queue. You can learn more about it in the [next section](#integrating-action-with-schemas).
 
+!!! note "Synchronous vs. Asynchronous Activities"
+    When defining activities, it is recommended to use synchronous functions by default. Asynchronous functions should only be used when you need to make asynchronous calls, for example, when interacting with an external API using `aiohttp`. Using blocking synchronous functions in an asynchronous activity can block the event loop and prevent other tasks from running.
```
ka-sarthak, line 144: Async functions can have a broader use case, for example, for reading large files from volumes using `aiofiles` or functions like gather from `asyncio`.

ka-sarthak (approved):
Looks great, thanks! Approving with a minor comment.

docs/howto/plugins/types/actions.md
blueraft: > Ordinary local file IO is blocking, and cannot easily and portably be made asynchronous. This means doing file IO may interfere with asyncio applications, which shouldn't block the executing thread. aiofiles helps with this by introducing asynchronous versions of files that support delegating operations to a separate thread pool.

`aiofiles` description :D If you are already using some other async functions in the same activity that's used to read files , `aiofiles` is fine to use, otherwise use the sync operation

PR #182, opened 2025-11-25
Architecture update

lauri-codes:
I updated our architecture image and text. Also added a section in the writing guide about images.


JFRudzinski (approved):
Great addition to the writing guide!

I added the "announcement" label so that the merge appears in Discord... maybe not the most important, but I would like to make docs announcements regularly, at least till people get used to the channel. That being said, feel free to remove the label before merge if you don't want this to be announced.

PR #186, opened 2025-12-01
Plugin navigation

JFRudzinski:
## Add Automated NOMAD Plugin Registry

This PR implements an automated plugin registry documentation page that queries and displays all NOMAD plugins from the FAIRmat-NFDI and nomad-coe GitHub organizations.

### Features

- **Automated Data Collection**: Queries the NOMAD Example Oasis `nomad-plugins` app for plugin metadata 
- **Multi-Organization Support**: Aggregates plugins from FAIRmat-NFDI and nomad-coe
- **Comprehensive Metadata**: 
  - Plugin details: name, description, repository, GitHub stars
  - Deployment status: PyPI availability, NOMAD Central, Example Oasis
  - Entry point types, authors, maintainers, and dependencies
- **Visual Statistics**: Overview metrics and Mermaid pie chart for plugin type distribution
- **Monthly Automation**: GitHub Actions workflow regenerates the page on the 1st of each month

### Implementation

**New Files:**
- `scripts/generate_plugin_registry.py` - API query and markdown generation script
- `.github/workflows/update-plugin-registry.yml` - Monthly automation workflow
- `docs/examples/plugin_registry.md` - Generated registry page
- `scripts/README.md` - Script documentation

**Workflow:**
- Schedule: Monthly on the 1st at 00:00 UTC
- Triggers: Scheduled run, manual dispatch, or push to main (for script changes)
- Creates PR for scheduled runs to allow review before merge

**Manual regeneration:**
```bash
python scripts/generate_plugin_registry.py

### CODE REVIEW

**Comment by JFRudzinski** [comment]
@DanielYang59 Before I send to others for review, could you please assess my overall approach here, provide feedback on the code, especially the github workflow which I have not tested yet.

File: examples/api/authenticated.py [hunk 1]
```diff
@@ -8,12 +8,12 @@
 from nomad.client import Auth
 
 nomad_url = config.client.url
-user = 'yourusername'
-password = 'yourpassword'
+user = "yourusername"
```
DanielYang59: not sure about the decision here, but in [the `nomad-fair` repo we used single quote](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/develop/pyproject.toml?ref_type=heads#L144-146)?

scripts/README.md
```diff
@@ -0,0 +1,69 @@
...
+  - Plugin dependencies
+
+### Usage
+
+```bash
+# Run directly
+python scripts/generate_plugin_registry.py
+
+# Or with uv
+uv run --with requests python scripts/generate_plugin_registry.py
```
DanielYang59: I guess instead of explicitly declare 3rd party dependencies in both the script and README, we could [declare script dependencies](https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies) (the same for the following `### Dependencies` section)?

scripts/generate_plugin_registry.py
```diff
@@ -0,0 +1,558 @@
...
+            markdown.append(plugin["description"])
+            markdown.append("")
+
+        # Repository and metadata
+        if plugin["repository"]:
+            markdown.append(
+                f"**Repository:** [{plugin['repository']}]({plugin['repository']})"
+            )
+
+        markdown.append(f"**Owner:** {plugin['owner']}")
```
DanielYang59: I guess we might need a few line breakers here

Image

scripts/generate_plugin_registry.py
```diff
@@ -0,0 +1,558 @@
...
+    # Build the page
+    page_content = []
+    page_content.append("# NOMAD Plugin Registry")
+    page_content.append("")
+    orgs_list = ", ".join(GITHUB_ORGS)
+    page_content.append(
+        "This page contains information about all NOMAD plugins owned and maintained by "
+        f" the GitHub organizations: {orgs_list}. "
+        " The information is automatically updated monthly. "
+        f"**Last Updated:** {datetime.now(timezone.utc).strftime('%Y-%m-%d %H:%M UTC')}"
```
DanielYang59: Guess here we need line breakers as well

Image

.github/workflows/update-plugin-registry.yml
```diff
@@ -0,0 +1,75 @@
...
+
+on:
+  schedule:
+    # Run on the 1st of every month at 00:00 UTC
+    - cron: '0 0 1 * *'
+  workflow_dispatch:
+    # Allow manual triggering of the workflow
+  push:
+    branches:
+      - main
```
DanielYang59: A bit unsure what condition is supposed to trigger this workflow, currently it would also be triggered by pushing to `main` (what about `develop`?) along with the monthly schedule:
> Monthly Automation: GitHub Actions workflow regenerates the page on the 1st of each month

.github/workflows/update-plugin-registry.yml
```diff
@@ -0,0 +1,75 @@
...
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: ${{ env.UV_VERSION }}
+
+      - name: Generate Plugin Registry
+        run: |
+          uv run \
+            --with requests \
+            --with regex \
+            python scripts/generate_plugin_registry.py
```
DanielYang59: Also here, I guess it could be very fragile to code these dependencies in different places?

.github/workflows/update-plugin-registry.yml
```diff
@@ -0,0 +1,75 @@
...
+        if: steps.check_changes.outputs.changed == 'true'
+        run: |
+          git config --local user.email "github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git add docs/examples/plugin_registry.md
+          git commit -m "docs: update plugin registry [automated]"
+          git push
+
+      - name: Create Pull Request (if on schedule)
+        if: github.event_name == 'schedule' && steps.check_changes.outputs.changed == 'true'
```
DanielYang59: this seems inconsistent with other triggers? i.e. manual trigger/push to main doesn't seem to able to trigger this step?

.github/workflows/update-plugin-registry.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Check for changes
+        id: check_changes
+        run: |
+          if git diff --quiet docs/examples/plugin_registry.md; then
+            echo "changed=false" >> $GITHUB_OUTPUT
+          else
+            echo "changed=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Commit and push if changed
```
DanielYang59: I think this step might be redundant? we should either push to main, or create a PR, currently it looks like we're pushing to main (nothing to commit for PR) then create PR?

examples/api/authenticated.py
JFRudzinski: @lauri-codes do you have a preference about this?

In any case I will not change the settings here, if we decide to align with the nomad-fair settings, we should create an issue for it, or else this PR will be flooded with a ton of other formatting changes

JFRudzinski: ah I see now that my linter has already made a ton of changes automatically, in this case, I will add the setting as you suggested to match nomad-fair

scripts/generate_plugin_registry.py
JFRudzinski: thanks for the feedback, actually this one-liner was intentional, otherwise the page is extremely long. I prefer it to be more compact. But let's see what the others say when they review.

scripts/generate_plugin_registry.py
JFRudzinski: same as above here

.github/workflows/update-plugin-registry.yml
JFRudzinski: Good point! The intention is to run the workflow on a schedule (right now monthly, but could be weekly or whatever). I removed the workflow run upon pushes

.github/workflows/update-plugin-registry.yml
JFRudzinski: thank you for pointing this out. I now go with the approach of always creating a PR if any changes are present. Can you please check again?

.github/workflows/update-plugin-registry.yml
JFRudzinski: addressed by the above changes

.github/workflows/update-plugin-registry.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: |
+          if git diff --quiet docs/examples/plugin_registry.md; then
+            echo "changed=false" >> $GITHUB_OUTPUT
+          else
+            echo "changed=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create Pull Request
+        if: steps.check_changes.outputs.changed == 'true'
+        uses: peter-evans/create-pull-request@v5
```
DanielYang59, line 45: Perhaps should [use Dependabot to auto-update actions](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)

---

looks like there is a newer v8 version now: https://github.com/peter-evans/create-pull-request/tree/main?tab=readme-ov-file#usage

also there is v6 for checkout: https://github.com/actions/checkout

setup-uv has v7: https://github.com/astral-sh/setup-uv

`uv` has 0.9: https://github.com/astral-sh/uv

.github/workflows/update-plugin-registry.yml
DanielYang59: great, thanks for the change

.github/workflows/update-plugin-registry.yml
DanielYang59: No problem and thanks for the change, I don't see any issue with the workflow file now

examples/api/getting_started.py
```diff
@@ -1,27 +1,23 @@
-'''
+"""
 Demonstrates how to use requests for a simple query and archive access.
-'''
+"""
 
-import requests
 import json
+import requests
```
DanielYang59: A nitpicking comment, can you perhaps check your linter/formater config (i.e. why 3rd party dependencies is grouped with built-in)?

I believe in [nomad-fair we're using isort via ruff](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/develop/pyproject.toml?ref_type=heads#L113)
```bash
uvx ruff check . --select I001 # --fix
```

JFRudzinski: applied. I don't really understand the ruff settings, I assumed they were set by the ruff version + toml. Is that not the case? Is there somehow where we store the nomad ruff standards/best practices config?

.github/workflows/update-plugin-registry.yml
JFRudzinski: could you possibly tackle this in a separate PR?

DanielYang59: sure

examples/api/getting_started.py
DanielYang59: > I don't really understand the ruff settings, I assumed they were set by the ruff version + toml. Is that not the case?

I believe [this is correct](https://docs.astral.sh/ruff/configuration/). The catch here is `ruff` is both a linter (via `ruff check`) and formatter (via `ruff format`). here the `[tool.ruff.format]` section only configured the formatter, and the linter would use [the default rule set](https://docs.astral.sh/ruff/rules/) which doesn't include `isort` (so here it's manually selected):
> By default, Ruff enables Flake8's F rules, along with a subset of the E rules, omitting any stylistic rules that overlap with the use of a formatter, like ruff format or [Black](https://github.com/psf/black).

---

> Is there somehow where we store the nomad ruff standards/best practices config?

I guess the easiest for us might be to use [ruff default](https://docs.astral.sh/ruff/settings/) unless we have a strong reason, here `indent-style`, `skip-magic-trailing-comma` and `line-ending` are already using the default so I guess we could just remove them from config. 

Regarding quotes [ruff would use double quotes by default](https://docs.astral.sh/ruff/settings/#format_quote-style) so if we decide to use single quotes (and [PEP 8 doesn't make a recommendation](https://peps.python.org/pep-0008/#string-quotes)), this is likely the only format config we need:
> In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it.

lukaspie:
Overall, I like this approach and lot and it is a very good addition to the docs. The automation looks great as well. Couple suggestions / discussion points:

**Plugin overview** 
- Not sure how editable the table format is, but the plugin overview table could maybe be a bit smaller by adjusting the column widths (first and last two could be smaller and the others wider).

**Detailed Plugin Information** 
- Maybe instead of the complete link to the repo, we could just write `Repository | Docs` and make each of them a link. I think linking the docs would be great in any case.
- Is `stars` a good stat to display? I get that it is a rough measure for how "successful" a plugin is, but a lot of these plugins have 0-2 stars, which may make them look undesirable for users. But I don't have a better proxy for how widely a package is adapated.
- Not sure I like `authors` and `maintainers` as two separate categories. A lot of these plugins also use "The NOMAD authors", which kind of defeats the purpose. This may be a more general discussion point: shall we always use this generic "The NOMAD authors" or really the actual authors? The latter is of course helpful if you want to talk to the devs directly, but may be harder to maintain with people leaving, etc.
- For the list of entry points, we could give for each of them their type (e..g `bandstructurenormalizer (Normalizer)`, then we could skip the item `Plugin Types`. But either way is fine.

**General**
- I discussed this with @hampusnasstrom already, the check whether a plugin is available on a given deployment is broke. But this is more an issue for the nomad-plugins code. **EDIT**: saw there is an [issue ](https://github.com/FAIRmat-NFDI/nomad-plugins/issues/22) for this already 👍 
- I would prefer if the automated script was run in shorter intervals, weekly maybe. We are still developing a lot of new plugins, would be great if they showed up on this page early.
- Does it make sense to have a similar page for non-FAIRmat plugins? This may help to connect the community.

ahm531:
Thanks Joe! My feedback mainly concerns the visual structure and information hierarchy of the page.

**Placement within the documentation**
I would recommend to place this page under *References* section rather than under *domain-specific examples* section. 
From a user perspective, a section titled “domain-specific examples” suggests curated examples grouped by research area (e.g. synthesis, computation, experiments, use cases, etc.). This would be very nice to have, but I imagine it would require manual curation which I see as not the main aim of this page at the moment. 

**Information in the overview table**
The overview table currently contains a large number of columns, which requires horizontal scrolling. Since the scrollbar only appears at the very bottom of the table, the layout may initially look broken or incomplete to users.
To improve usability, I suggest reducing the table width by limiting it to the most essential information, so that it fits within a typical window size (to eliminate the need of the scrollbar). From my perspective, the most important fields for the overview are:
- plugin name
- description
- availability / deployment status (e.g. tags for PyPI, NOMAD Central, Example Oasis, aligned vertically)
- links (at least to the repository, and ideally also to documentation as suggested by Lukas)

I am a bit undecided about the usefulness of displaying stars at this stage, as many repositories currently have zero stars. If they are to be kept, I would suggest placing them next to the plugin name rather than as a separate column. The same applies to plugin types: they could appear as small tags under the plugin name, or potentially be omitted from the overview table altogether, since this information is already available in the detailed view.

For examples: here is one suggestion: 
image
  
**Access to the detailed information**
I think it would be very helpful if the detailed plugin information were linked from each table entry (for example via expandable rows or a dedicated details view). This would allow users to quickly scan the overview and then head down into the details of a specific plugin only when needed.
In addition, having these detailed informaion in the ### level expands the right navigation panel signficantly. I'm not sure that is really useful to actually navigate the page. 

Another suggestion is to merge the detailed information within the table in accordion style, which could look like this: 
image

ahm531 (commented):
Hey Joe,
I have put in my review as a general feedback in the comments part. 
Let me know if a detailed review is needed from my side.

JFRudzinski:
> ar as small tags under the plugin name, or potentially be omitted from the overview table altogether, since this information is already available in the detailed vie

Thanks for the suggestions @ahm531 , did you create this example "manually" or did you edit the generation script? In the former case, can you supply some raw md for me? In that latter case, could you either share the script or even create a PR of your branch with this as the target branch?

ahm531:
@JFRudzinski I did add these examples manually.
The attached .md file [test_plugin_registry.md](https://github.com/user-attachments/files/24766746/test_plugin_registry.md) contains both cases.

Example 1: implemented using pure Markdown

Example 2: a collapsible section. I couldn’t achieve this with pure Markdown in our setup, so I used HTML for that part.

JFRudzinski:
Thank you @lukaspie and @ahm531 for the feedback. Below I address individual comments and note changes so far:

From @lukaspie:

> Plugin overview

> - Not sure how editable the table format is, but the plugin overview table could maybe be a bit smaller by adjusting the > column widths (first and last two could be smaller and the others wider).

Done. (According to Ahmed's suggestion)

> Detailed Plugin Information

> - Maybe instead of the complete link to the repo, we could just write Repository | Docs and make each of them a link. I think linking the docs would be great in any case.

Done.

> - Is stars a good stat to display? I get that it is a rough measure for how "successful" a plugin is, but a lot of these plugins have 0-2 stars, which may make them look undesirable for users. But I don't have a better proxy for how widely a package is adapated.

I don't really have a strong opinion myself on this myself. I see that at the moment they are not very useful, but perhaps in the future they are more indicative of usage? I left them now for further comments from others.

> - Not sure I like authors and maintainers as two separate categories. A lot of these plugins also use "The NOMAD authors", which kind of defeats the purpose. This may be a more general discussion point: shall we always use this generic "The NOMAD authors" or really the actual authors? The latter is of course helpful if you want to talk to the devs directly, but may be harder to maintain with people leaving, etc.

Can you explain why you don't like 2 separate categories? This is well-defined in GitHub. For me the author list is for acknowledgement. The maintainer list for contact. It is also mentioned at the top of the page that FAIRmat owns and maintains all these plugins. Perhaps that should be made clearer.

Usage of "The NOMAD Authors" appears to be largely used by nomad-coe plugins which are or will eventually be archived. 

- [ ] We will need to add a check for archived repos and perhaps just list them concisely at the bottom of the page or something.

The issue of maintaining author lists is independent of this page. But I think it is something that we want to do, at least while FAIRmat exists, as it is a major way we can acknowledge our co-workers work, through publications of repos on Zenodo.

> - For the list of entry points, we could give for each of them their type (e..g bandstructurenormalizer (Normalizer), then we could skip the item Plugin Types. But either way is fine.

I am fine with this but some plugins have many entry points which will blow up the text a little. I like the way Ahmed did it cause the Plugin Types are in the overview, and the detailed list only in the drop down. Also, often the type is also in the entry point name, as in the example you gave. I am open to more opinions on this though.

> General

> - I discussed this with @hampusnasstrom already, the check whether a plugin is available on a given deployment is broke. But this is more an issue for the nomad-plugins code. EDIT: saw there is an https://github.com/FAIRmat-NFDI/nomad-plugins/issues/22for this already 👍

:+1: 

> - I would prefer if the automated script was run in shorter intervals, weekly maybe. We are still developing a lot of new plugins, would be great if they showed up on this page early.

Done. (There is a check that only opens a PR if changes are found, so this should have no negative side effect).

> - Does it make sense to have a similar page for non-FAIRmat plugins? This may help to connect the community.

Yes, but let's finalize the format first.

- [x] Create an analogous page for external plugins

From @ahm531 

I took the format suggestions from `test_plugin_registry.md` directly. If there is some other consideration that I missed, please let me know.

JFRudzinski:
Oh .. The location of the page let's discuss later

lukaspie:
> From @lukaspie:
 
> > * Not sure I like authors and maintainers as two separate categories. A lot of these plugins also use "The NOMAD authors", which kind of defeats the purpose. This may be a more general discussion point: shall we always use this generic "The NOMAD authors" or really the actual authors? The latter is of course helpful if you want to talk to the devs directly, but may be harder to maintain with people leaving, etc.
> 
> Can you explain why you don't like 2 separate categories? This is well-defined in GitHub. For me the author list is for acknowledgement. The maintainer list for contact. It is also mentioned at the top of the page that FAIRmat owns and maintains all these plugins. Perhaps that should be made clearer.
> Usage of "The NOMAD Authors" appears to be largely used by nomad-coe plugins which are or will eventually be archived.
> The issue of maintaining author lists is independent of this page. But I think it is something that we want to do, at least while FAIRmat exists, as it is a major way we can acknowledge our co-workers work, through publications of repos on Zenodo.
> 
I think it's fine to have both categories, as long as we agree to fill these correctly. That means to me that we do not use "The NOMAD authors", but highlight the actual people who did the work and who are the maintainers. For now, the categories are pretty redundant if they just contain these generic placeholders in both categories. But okay, let's keep and make it a separate point to acknowdledge properly. I think the main issue is the inconsistent use of authors and maintainers in `pyproject.toml` (where often just  "The NOMAD Authors" is used) and in the CITATION.cff for Zenodo, where often the proper list is given.

On a separate note, the maintainers category may be tricky because these get outdated with coworkers leaving.

> * [ ]  We will need to add a check for archived repos and perhaps just list them concisely at the bottom of the page or something.

agree 

> > * For the list of entry points, we could give for each of them their type (e..g bandstructurenormalizer (Normalizer), then we could skip the item Plugin Types. But either way is fine.
> 
> I am fine with this but some plugins have many entry points which will blow up the text a little. I like the way Ahmed did it cause the Plugin Types are in the overview, and the detailed list only in the drop down. Also, often the type is also in the entry point name, as in the example you gave. I am open to more opinions on this though.
> 

agree, I like Ahmed's solution

lukaspie:
One more thing that I forgot (maybe it was already discussed, if so, please ignore): would it be possible to have at the top of the plugins overview the possiblity to filter by entry point? So e.g., all NORTH tools or all Parsers could be shown?

Or do we let the `NOMAD plugins app` handle that?

lauri-codes:
I do think that having the list of plugins cleanly in very clear webpage is a good idea. To me this also feels much more natural than having a separate app for plugins. Although functionality-wise the app is great, it is very hard to justify that approach because it is too tighlty coupled to a deployment and requires quite a bit of mental overhead to understand for new people ("I'm now in a NOMAD deployment, which has a NOMAD App that showcases NOMAD plugins." vs. "I'm now reading the NOMAD docs and looking at a list of plugins").

Functionality-wise and visually this can certainly be improved (I would ask Berfin about the visuals), but I think there are more fundamental questions:

- Should the list include also third-party plugins (I think so, maybe in a separate view or at least so that they can be easily also fitlered out).
- Should this completely replace the plugins app (I think so)
- Should this table be tied to the docs, or do we put this into the nomad-lab.eu homepage? The homepage is absolutely not tied to any installation (the docs are), and I think nomad-lab.eu is really the page for decision-makers: people who want to learn about the infrastructure.
- In the deployment-specific docs we could still pull information about the plugins that are installed to that specific deployment if that is something we wish to do.

JFRudzinski:
> I do think that having the list of plugins cleanly in very clear webpage is a good idea. To me this also feels much more natural than having a separate app for plugins. Although functionality-wise the app is great, it is very hard to justify that approach because it is too tighlty coupled to a deployment and requires quite a bit of mental overhead to understand for new people ("I'm now in a NOMAD deployment, which has a NOMAD App that showcases NOMAD plugins." vs. "I'm now reading the NOMAD docs and looking at a list of plugins").
> 
> Functionality-wise and visually this can certainly be improved (I would ask Berfin about the visuals), but I think there are more fundamental questions:
> 
> * Should the list include also third-party plugins (I think so, maybe in a separate view or at least so that they can be easily also fitlered out).
> * Should this completely replace the plugins app (I think so)
> * Should this table be tied to the docs, or do we put this into the nomad-lab.eu homepage? The homepage is absolutely not tied to any installation (the docs are), and I think nomad-lab.eu is really the page for decision-makers: people who want to learn about the infrastructure.
> * In the deployment-specific docs we could still pull information about the plugins that are installed to that specific deployment if that is something we wish to do.

Based on these comments, I think there are some fundamental design choices to discuss. I will arrange a dedicated meeting for this.

lukaspie:
I just came across this, we should probably remove it as well or link to the plugin navigation page instead: https://nomad-lab.eu/prod/v1/docs/howto/plugins/types/schema_packages.html#schema-packages-developed-by-fairmat

JFRudzinski:
@DanielYang59 I merged develop into this branch, and now I get a ton of warnings when I do mkdocs serve which I guess are coming from the changes from removing the example data maybe. Have you seen this? Can you check it for me?

JFRudzinski:
Current status:

plugin_registry

JFRudzinski:
TODOs:
- [ ] extend plugin metadata for further filtering (archived, categories)
- [ ] extend filter to cover domain-specific examples and then (if successful) remove examples section entirely

DanielYang59:
> @DanielYang59 I merged develop into this branch, and now I get a ton of warnings when I do mkdocs serve which I guess are coming from the changes from removing the example data maybe. Have you seen this? Can you check it for me?

Can you show me the warnings you got? I only saw some deprecation warnings from fastapi side but was fixed in https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/2836
```
INFO    -  FastAPIDeprecationWarning: `example` has been deprecated, please use `examples` instead
             File "/home/yang/nomad/nomad-docs/.venv/lib/python3.12/site-packages/fastapi/params.py", line 521, in __init__
               warnings.warn(
             File "/home/yang/nomad/nomad-docs/.venv/lib/python3.12/site-packages/nomad/app/v1/models/models.py", line 380, in
               class WithQuery(BaseModel):
INFO    -  FastAPIDeprecationWarning: `example` has been deprecated, please use `examples` instead
             File "/home/yang/nomad/nomad-docs/.venv/lib/python3.12/site-packages/fastapi/params.py", line 521, in __init__
               warnings.warn(
             File "/home/yang/nomad/nomad-docs/.venv/lib/python3.12/site-packages/nomad/app/v1/models/models.py", line 1134, in
               class WithQueryAndPagination(WithQuery):
INFO    -  FastAPIDeprecationWarning: `example` has been deprecated, please use `examples` instead
             File "/home/yang/nomad/nomad-docs/.venv/lib/python3.12/site-packages/fastapi/params.py", line 521, in __init__
               warnings.warn(
             File "/home/yang/nomad/nomad-docs/.venv/lib/python3.12/site-packages/nomad/app/v1/models/models.py", line 1140, in
               class Metadata(WithQueryAndPagination):
```

JFRudzinski:
yeah same:

```
INFO    -  The following pages exist in the docs directory, but are not included in
           the "nav" configuration:
             - 404.md
             - aitoolkit.md
             - todo.md
             - tutorials_template.md
             - writing_guide.md
             - examples/experiment_data/nexus.md
             - howto/manage/program/graph-api/basics.md
             - tutorial/access_api.md
             - tutorial/nomad_repo.md
INFO    -  FastAPIDeprecationWarning: `example` has been deprecated, please use
           `examples` instead
             File
           "/home/jfrudzinski/work/soft/plugins/nomad-docs/.venv/lib/python3.12/site-packages/fastapi/params.py",
           line 521, in __init__
               warnings.warn(
             File
           "/home/jfrudzinski/work/soft/plugins/nomad-docs/.venv/lib/python3.12/site-packages/nomad/app/v1/models/models.py",
           line 380, in
               class WithQuery(BaseModel):
INFO    -  FastAPIDeprecationWarning: `example` has been deprecated, please use
           `examples` instead
             File
           "/home/jfrudzinski/work/soft/plugins/nomad-docs/.venv/lib/python3.12/site-packages/fastapi/params.py",
           line 521, in __init__
               warnings.warn(
             File
           "/home/jfrudzinski/work/soft/plugins/nomad-docs/.venv/lib/python3.12/site-packages/nomad/app/v1/models/models.py",
           line 1134, in
               class WithQueryAndPagination(WithQuery):
INFO    -  FastAPIDeprecationWarning: `example` has been deprecated, please use
           `examples` instead
             File
           "/home/jfrudzinski/work/soft/plugins/nomad-docs/.venv/lib/python3.12/site-packages/fastapi/params.py",
           line 521, in __init__
               warnings.warn(
             File
           "/home/jfrudzinski/work/soft/plugins/nomad-docs/.venv/lib/python3.12/site-packages/nomad/app/v1/models/models.py",
           line 1140, in
               class Metadata(WithQueryAndPagination):
```

but also these files not in nav. I guess maybe I messed up the merge?

DanielYang59:
I think there're two separate warnings here:
- `The following pages exist in the docs directory, but are not included the "nav" configuration:` for some markdowns in the docs directory but not in the nav section of `mkdoc.yml`
- The rest `FastAPIDeprecationWarning` is from `nomad-lab` side and [has been fixed](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/2836), the lock files was not yet updated (i would do this later) https://github.com/FAIRmat-NFDI/nomad-docs/blob/ec2ba55ca02af3b33fe97e75627e5a6252dad79e/uv.lock#L1837-L1838

You could run `uv sync --upgrade` and the deprecation would be gone

g-berfinguener (approved):
Hello Joe, Went through the preview, here is my feedback.

Urgent, before deploying

- The Containing and Owner filters are multi-select listboxes an the options are cut off by the scrollbar, only four of six visible at a time. Checkboxes can be the fix.
- There is no way to search the plugin list itself. A search field in the filter panel would help in my opinion but up to you ofcourse.
- The amber warning banner shows in the default state, so the first thing every visitor reads is a warning. I would either neutralise the wording and the color and move the FAIRmat vs community distinction to a per-row badge.
- Owner has both "Other" and "Non-FAIRmat" as options, which can be confusing I think.
- Rows with zero stars still show the star icon with 0, which is most of the table. Hide stars when the count is 0.
- The full width View Details bar under every row doubles the table height. Making the row itself clickable with a chevron would take much less space.

Non-urgent, for a follow-up issue

- The Deployment column is a dash for most rows, could move into the expanded details instead.
- Sorting could live in the column headers rather than in the filter panel. Or, if we go with the chart-as-filter approach, sorting could sit right below the charts alongside the active filter chips.

JFRudzinski:
closed as schema and workflow is moved to https://github.com/FAIRmat-NFDI/nomad-plugins, with registry moved to https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-lab-homepage/

PR #193, opened 2026-01-02
Update nomadsims url

JFRudzinski:


JFRudzinski:
checked out from wrong branch

PR #204, opened 2026-01-23
Fix broken link in docs

DanielYang59:


DanielYang59:
here is another broken link, not sure what should be the replacement: https://github.com/FAIRmat-NFDI/nomad-docs/blob/develop/examples/data/eln/README.md?plain=1#L12

JFRudzinski:
> here is another broken link, not sure what should be the replacement: https://github.com/FAIRmat-NFDI/nomad-docs/blob/develop/examples/data/eln/README.md?plain=1#L12

Can you tell me where/how this is being used? I am not so familiar with all the stuff under `data/`, is it being imported programmatically somewhere?

JFRudzinski:
@lauri-codes do you know anything about the history of the content in `examples/`, e.g., [THIS](https://github.com/FAIRmat-NFDI/nomad-docs/blob/11cf6a6995c92e393a37ff03e4aca7a22a8752bb/examples/data/eln/README.md?plain=1#L12C63-L12C109)

it seems like this is not actually being used. It would be good to remove stuff if it is old and/or we have no intent of using it in the future

lauri-codes:
The only reason to keep examples is if they are either used by our  Python source code (src/nomad_docs) or then are being read in as file snippets into the markdown files (in which case they would also ideally be tested by a test in `tests`.

Any files that are not currently being referenced or used can be dropped in my opinion.

PR #211, opened 2026-02-03
Update to scan nested md references

### LINKED ISSUE #210: test_assets.py marks images referenced from nested docs as  unused
`test_assets.py` may report images/data as unused when they are referenced from md files in nested folders (e.g. `docs/tutorial/eln/*.md` linking via `../images/...`). This leads to false positives for section-level assets.

siamakn:
Fixes the test so it finds image/data references from nested docs pages.


blueraft (approved):
Thank you

PR #213, opened 2026-02-05
Remove unused assets in `examples/`, relocate `ops/`

### LINKED ISSUE #208: Clean up unused `examples`
https://github.com/FAIRmat-NFDI/nomad-docs/pull/204#issuecomment-3839549807

Currently it seems some files in `examples` directory might not be used, should drop them if so.

DanielYang59:
Fix https://github.com/FAIRmat-NFDI/nomad-docs/issues/208


DanielYang59:
Hi @JFRudzinski @lauri-codes  I believe this is ready for review. This really "removed all assets that are currently not used in docs", but still:
- Would some of them depend on another? for some in `examples/data/eln` only `schema.archive.yaml` is "used" but there are also README and other things, not sure if such removal would make it un-readable(missing README or other metadata) or un-trackable (some could be processed data from other raw data)?
- Most files are used only once (with the exception of `examples/archive/` and `examples/metainfo/data_frames.py` if I remember correctly), not sure if we should just merge them into corresponding markdowns?

lauri-codes (approved):

PR #232, opened 2026-02-17
Update the API request according to the new API

RubelMozumder:
According to the new API implementation, to get an `Auth` token, one has to make a `POST` request rather than a GET.

@DanielYang59, would you please verify this change?


DanielYang59 (approved):
thanks yes it's correct (I thought I have updated the docs completely)

PR #235, opened 2026-02-25
Add docs on tuning temporal workers

blueraft:


docs/howto/oasis/configure.md
```diff
@@ -202,44 +202,110 @@ Linux), you might run into problems with processing large uploads. If the NOMAD
...
+
+---
+
+### 3. Resource Management & Guardrails
+
+To keep your cluster healthy, you must combine infrastructure constraints with NOMAD's application-aware guardrails.
+
+* **Hard Limits (Infrastructure):** Your Kubernetes or Docker pod resource limits (`resources.limits.memory` and `resources.limits.cpu`) are the ultimate guardrails. They protect your host node from being completely consumed by a runaway worker.
+* **Soft Limits (NOMAD `WorkerConfig`):** Settings like `target_memory_usage` and `target_cpu_usage` act as an early-warning system. They allow the worker to gracefully pause accepting new tasks from the queue *before* the container hits the hard Kubernetes limit. This prevents aggressive and disruptive Out-Of-Memory (OOM) kills.
```
hampusnasstrom: Are these limits percentages of the hard limits or the total system resources?

docs/howto/oasis/configure.md
```diff
@@ -202,44 +202,113 @@ Linux), you might run into problems with processing large uploads. If the NOMAD
...
+```yaml
+temporal:
+  internal_worker:
+    # Stop accepting new tasks if container CPU hits 80%
+    target_cpu_usage: 0.8
+    # Stop accepting new tasks if container RAM hits 80%
+    target_memory_usage: 0.8
+```
+
+---
```
lauri-codes: Swap the order: 3 -> 2

docs/howto/oasis/configure.md
```diff
@@ -202,44 +202,113 @@ Linux), you might run into problems with processing large uploads. If the NOMAD
...
+temporal:
+  internal_worker:
+    # Number of Python processes running in this container
+    pool_size: 4
+    # Restart the process after 100 tasks to clear memory leaks
+    max_tasks_per_child: 100
 ```
 
-Each replica runs as an independent worker process. Docker will handle restarting and load balancing between them.
+* **Memory impact:** Because these are distinct processes, a single container with a `pool_size` of 10 will consume roughly the same amount of memory as 10 separate containers with a `pool_size` of 1.
```
lauri-codes: This needs some more details, and also recommending to set this to the number of cores is not always ideal.

I would instead recommend here our current default: pool_size=1, and then say that depending on the type of processing that is being done, this can be increased up to the number of CPU cores. Then just mention that this typically only makes sense if your processing does not fully utilize your available cpu cores (your process are not already bypassing GIL e.g. by doing heavy numpy calculations, or you have not already scaled the rpocessing using replicas) and if you have enough memory (mention the memory impact).

docs/howto/oasis/configure.md
```diff
@@ -202,44 +202,113 @@ Linux), you might run into problems with processing large uploads. If the NOMAD
...
+Workloads involving large datasets or trajectories that must be loaded entirely into RAM.
+
+* **Behavior:** High risk of sudden Out-Of-Memory (OOM) crashes.
+* **How to tune:** This scenario requires strict isolation. Favor higher replica counts with very low `pool_size` limits (even a `pool_size: 1`). This ensures that if a massive dataset causes an unavoidable OOM spike, it only kills one isolated replica rather than a pooled worker that is concurrently processing other users' data. Rely heavily on strict Kubernetes memory limits combined with NOMAD's `target_memory_usage: 0.7` to reject tasks gracefully when RAM gets tight.
+
+There are several ways to control resource usage and improve performance:
+
+- Increase or decrease the number of worker **replicas** (recommended)
+- Adjust CPU and memory limits in Docker
+- (Optionally) Increase the number of worker **processes** on the Python side
```
lauri-codes: This feels out of place: remove?

docs/howto/oasis/configure.md
lauri-codes: This should either be removed, or them moved to the "Hard limits" section

lauri-codes (commented):
I like it a lot, some comments.

docs/howto/oasis/configure.md
blueraft: Done

lauri-codes (approved):

Issue #236, opened 2026-02-26
configuration for backend scopes, and usage from user side


DanielYang59:
- https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/2747
- https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/2854

@hampusnasstrom I heard from @lauri-codes that you're working on this? If you haven't started yet maybe I could start first and let you review/comment?

hampusnasstrom:
@DanielYang59 that would be great, thank you! I think one thing which would be good is an example in the end. But just request a review from me when you have something for me to look at :)

PR #241, opened 2026-03-06
Add docs for auth configs 

DanielYang59:
- [x] Add scopes docs, close https://github.com/FAIRmat-NFDI/nomad-docs/issues/236
- [x] Clarify layers of auth


docs/howto/oasis/configure.md
```diff
@@ -275,39 +275,217 @@ See also the [docker-compose documentation](https://docs.docker.com/compose/comp
...
+
+For example, if an endpoint requires `uploads:write` and the token only has
+`uploads:read`, the request will be rejected.
+
+#### Keycloak access tokens
+
+Keycloak access tokens authenticate a user with their full account permissions.
+At the backend level, they are treated as having the full set of user scopes.
+
+#### Personal Access Tokens (PATs)
```
DanielYang59: this would link to https://github.com/FAIRmat-NFDI/nomad-docs/pull/247 later

DanielYang59:
@hampusnasstrom I just finished the first draft, maybe you could have a look and point out what is unclear/missing? thanks

docs/howto/oasis/configure.md
```diff
@@ -275,39 +275,257 @@ See also the [docker-compose documentation](https://docs.docker.com/compose/comp
...
+
+```text
+*:read   # read-only access to all resources
+*:*      # full access
+```
+
+!!! note
+    Partial wildcard patterns such as `u*:read` are not supported.
+
+All available scopes are defined in the `nomad.auth.scopes.Scope` enum.
```
hampusnasstrom: Is it possible to include these in the reference docs and have a link here? Or maybe at least have a link to the file on Gitlab?

docs/howto/oasis/configure.md
```diff
@@ -275,39 +275,257 @@ See also the [docker-compose documentation](https://docs.docker.com/compose/comp
...
 ```yaml
-oasis:
-    allowed_users:
-        - user1@gmail.com
-        - user2@gmail.com
+auth:
+  require_authentication: true
+  authorized_users:
+    - alice@example.com
+    - bob@example.com
```
hampusnasstrom, line 445: I would add one more example as I think it's a very common use case. I'm not sure about the defaults though.

```suggestion
    - bob@example.com

  Restricted Oasis: Anonymous and users not in whitelist can only read (published) data.
  ```yaml
  auth:
    unauthorized_user_scopes:
      include:
        - "*:read"
    authorized_users:
      - alice@example.com
      - bob@example.com
```

docs/howto/oasis/configure.md
```diff
@@ -275,39 +275,257 @@ See also the [docker-compose documentation](https://docs.docker.com/compose/comp
...
+
+```yaml
+auth:
+  require_authentication: false
+  unauthenticated_user_scopes:
+    include:
+      - "*:read"
+```
+
+Private Oasis: login required.
```
hampusnasstrom: This name and the config example is a bit misleading I think. The vast majority of Oases use the central keycloak. For those Oases this config doesn't really make the Oasis "private" as anyone can make an account on nomad-lab. I would remove this example and rename the next one with the whitelist to "Private".

docs/howto/oasis/configure.md
```diff
@@ -275,39 +275,257 @@ See also the [docker-compose documentation](https://docs.docker.com/compose/comp
...
-    require_authentication: true
+auth:
+  require_authentication: true
 ```
 
-Lastly, we offer a simple white-list mechanism. As the Oasis administrator you provide a
-list of accounts as part of your Oasis configuration. To use the Oasis, all users have to
-be logged in and be on your white list of allowed users. To enable white-listing, you
-can provide a list of NOMAD account email addresses in your `nomad.yaml` like this:
+Restricted Oasis: Only specific users in whitelist can access the system.
```
hampusnasstrom: See comment above
```suggestion
Private Oasis: Only specific users in whitelist can access the system.
```

docs/howto/oasis/configure.md
```diff
@@ -275,39 +275,257 @@ See also the [docker-compose documentation](https://docs.docker.com/compose/comp
...
+When this option is enabled, all requests must include a valid authentication token.
+Otherwise the API will return:
+
+```text
+HTTP 401 Unauthorized
+```
+
+### Scope-based authorization
+
+After authentication, NOMAD determines **which actions the user is allowed to perform**s
```
hampusnasstrom:
```suggestion
After authentication, NOMAD determines **which actions the user is allowed to perform**
```

hampusnasstrom (changes requested):
Really nice documentation Haoyu! I left some small comments but the two major things are:

1. I think we are missing the reference docs for this auth config if I'm not mistaken? This should be added with all the default values and so on and then linked somewhere in this section.
2. I'm not sure if this "How-to" guide really follows the Diataxis principles. There is a lot of (very nicely written and informative!) explanation. However, reading the rest of this "How-to" guide it sort of fits so I'm not sure. @lauri-codes what do you think? Does it make sense to move something to the explanation or should we just leave it here?

docs/howto/oasis/configure.md
DanielYang59: oh I see, I didn't realize we have a reference section at all. would update

docs/howto/oasis/configure.md
```diff
@@ -275,39 +275,260 @@ See also the [docker-compose documentation](https://docs.docker.com/compose/comp
...
-through a VPN.
+Access control can therefore be configured on several levels:
 
-Secondly, you can require authentication for all sensitive endpoints by enabling
-the global `require_authentication` flag in your configuration:
+1. Network level — restrict access via firewall, VPN, or private network.
+2. Authentication level — require users to log in before accessing the API.
+3. Authorization level (scopes) — control which operations users are allowed to perform after login.
+
+The following diagram summarizes how access is evaluated. [Administrator-configurable settings](../../reference/config.md#auth) are highlighted in red.
```
DanielYang59: > I think we are missing the reference docs for this auth config if I'm not mistaken? This should be added with all the default values and so on and then linked somewhere in this section.

added here

docs/howto/oasis/configure.md
DanielYang59: I tried but didn't come up with a too good solution here:
- either we create a [macro](https://github.com/FAIRmat-NFDI/nomad-docs/blob/develop/src/nomad_docs/__init__.py#L151) to dump this `enum` but it might flood the documentation page
- linking to the file is a bit tricky because we cannot directly link to the `develop` branch (which is likely to be out of sync with doc version), but I think I could write a macro to extract current `nomad-lab` version in `docs` and link to corresponding tag (e.g. `https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/v1.4.1/nomad/auth/scopes.py`)? do you think we should do this? @lauri-codes

lauri-codes:
Thanks Haoyu!

This is already **too** comprehensive, and mixes too many things together. So we just need to pull stuff from this PR to others. It now contains all of these topics:

- Overview of authentication vs authorization + auth decision tree
- Contains a summary of the token types
- Documentation about scopes
- How to restrict Oasis access using config variables

We should only focus on the last item in this list for now, and do the rest on other PRs and other locations in the docs. I think you can simply have the following subsections, which should very short:

```md
# Restricting access to your Oasis

Access control to an Oasis can be configured on the following three levels levels:

 1. Network level — restrict access via firewall, VPN, or private network.
 2. Authentication level — require users to log in before accessing the API.
 3. Authorization level — control which users and which operations are allowed after login.
 
 (If we want to have a diagram about auth, this should go into the "explanation" section and we could link to it here).
 
The network level is not controlled by the NOMAD software itself, so the following sections focus on the two remaining levels: authenticatiton and authorization.

## Requiring authentication
(Don't talk here about different ways to authenticate, just focus on how you can require authentication using the config)

## Unauthenticated user permissions
(This looks good already)

## Restricting access to specific users
(Only focus on the whitelisting)

## Unauthorized user permissions
(Only focus on the unauthorized users config)

## Example configurations
(Looks good already)
```

Most importantly this part of the doc should not explain the scopes: we need a separate page for them. Once this is in place (I think it will come together with the PAT docs), you can link to there.

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,166 @@ services:
...
+```
+
+This allows anonymous users to browse published data but prevents modifications.
+
+#### Configure non-whitelisted user permissions
+
+Users who successfully authenticate but are not in the `authorized_users` whitelist
+are handled according to the `reject_unauthorized_users` setting.
+
+When enabled (`reject_unauthorized_users: true`), non-whitelisted users will be rejected with:
```
DanielYang59: @lauri-codes I think the terminology "unauthorized" in config (`reject_unauthorized_users`, `unauthorized_user_scopes` and `authorized_users`) has been confusing to me... "unauthorized" seems to imply full rejection instead of still having partial access, maybe we should rename them to "non-whitelisted" (or something else)?

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,166 @@ services:
...
+```
+
+When this option is enabled, all requests must include a valid authentication token.
+Otherwise the API will return:
+
+```text
+HTTP 401 Unauthorized
+```
+
+#### Restricting access to specific users
```
DanielYang59, line 310: maybe we need better wording here because  `authorized_users` alone isn't enough to "restrict access to specific users"... (require_authentication is false by default)

But if we add `require_authentication=true` it would overlap with the configuration example...

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,166 @@ services:
...
+Restricted Oasis: Only specific users in the whitelist can access the system.
 
 ```yaml
-oasis:
-    allowed_users:
-        - user1@gmail.com
-        - user2@gmail.com
+auth:
+  require_authentication: true
+  reject_unauthorized_users: true
```
DanielYang59, line 442: `reject_unauthorized_users` is true by default, but maybe having it here is better for demonstration purpose?

DanielYang59:
@hampusnasstrom maybe you could have another look? Thanks

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,159 @@ services:
...
+When this option is enabled, all requests must include a valid authentication token.
+Otherwise the API will return:
+
+```text
+HTTP 401 Unauthorized
+```
+
+#### Restricting access to specific users
+
+Administrators can also restrict access to an explicit whitelist of users.
```
DanielYang59: [Hampus mentioned](https://discord.com/channels/1201445470485106719/1482049485239488622) in our currently docs we only mentioned email:
> To enable white-listing, you can provide a list of NOMAD account email addresses in your nomad.yaml like this:

But as far as I know username and email should work the same, maybe we should mention this?

DanielYang59: In our config doc:
> A list of usernames or user account emails that are authorized to access this NOMAD deployment. If not specified, all users recognized by the Keycloak instance are allowed.

I also added a username as example here (current example only has email, if people don't read config docs, they would easily misunderstand)

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,160 @@ services:
...
-through a VPN.
+Access control can therefore be configured on several levels:
 
-Secondly, you can require authentication for all sensitive endpoints by enabling
-the global `require_authentication` flag in your configuration:
+1. Network level — restrict access via firewall, VPN, or private network.
+2. Authentication level — require users to log in before accessing the API.
+3. Authorization level (scopes) — control which operations users are allowed to perform after login.
+
+We included a diagram summarizing how access is evaluated for better understanding of the following configs.
```
lauri-codes: Now with the other PR done, modify this to say:

"You can learn more about authentication and authorization in our explanation-section."

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,160 @@ services:
...
+
+<!-- TODO: add link to diagram -->
```
lauri-codes: Remove todo

docs/howto/oasis/configure.md
lauri-codes: To make this absolutely clear, say: "The `authorized_users` is a list of usernames or user emails. If specified, only these users are considered to be fully authorized for access."

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,160 @@ services:
...
+Semantics:
+
+- `include` defines the baseline scopes
+- `exclude` removes scopes from that baseline (with higher precedence than `include`)
+- wildcards are supported
+
+!!! note
+    Partial wildcard patterns such as `u*:read` are **not** supported.
+
+All available scopes are defined in the `nomad.auth.scopes.Scope` enum.
```
lauri-codes: Add a link to the new tokens doc which contains the table.

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,160 @@ services:
...
+
+When an API endpoint is called, the backend checks whether the user has the required scopes.
+If required scopes are missing, the API returns:
+
+```text
+HTTP 403 Forbidden
+Missing scopes: [...]
+```
+
+#### Configure anonymous user permissions
```
lauri-codes: Rename this to "Configure unauthenticated user permissions". This matches the config variable name.

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,160 @@ services:
...
-Currently there are three ways to restrict access to your Oasis. First, you do not
-expose the Oasis to the public internet, e.g. you make it only available on an intra-net or
-through a VPN.
+Access control can therefore be configured on several levels:
 
-Secondly, you can require authentication for all sensitive endpoints by enabling
-the global `require_authentication` flag in your configuration:
+1. Network level — restrict access via firewall, VPN, or private network.
+2. Authentication level — require users to log in before accessing the API.
+3. Authorization level (scopes) — control which operations users are allowed to perform after login.
```
lauri-codes: Remove "scopes" from here: Authorization includes more than that.

docs/howto/oasis/configure.md
lauri-codes: Also we should link to the `auth` section in our config. Could you add at the end of this paragraph something like:

"The authentication and authorization settings for individual NOMAD deployments are configurable through the [auth](link here) configuration section in `nomad.yaml` and the following sections demonstrate its usage."

lauri-codes (changes requested):

docs/howto/oasis/configure.md
DanielYang59: added, also mentioned this list is case insensitive

docs/howto/oasis/configure.md
hampusnasstrom: @DanielYang59 Could you please add this example instead?

docs/howto/oasis/configure.md
```diff
@@ -273,41 +273,178 @@ services:
...
+the Oasis but only with restricted access, configurable via:
+
+```yaml
+auth:
+  unauthorized_user_scopes:
+    include:
+      - "*:read"
+```
+
+### Example configurations
```
DanielYang59, line 401: @hampusnasstrom @lauri-codes I just tried to clarify the examples here a bit, maybe you could check if there's any improvement needed.

- Add a new example config, https://github.com/FAIRmat-NFDI/nomad-docs/pull/241#discussion_r2929513955, I believe that example is missing `reject_unauthorized_users=False`, can you double check? @hampusnasstrom 
- add more descriptive name instead of just "restricted oasis"
- sorted the example config order from "open" to "restricted"
- in all these examples, the related "default config" is still repeated for better demonstration?

hampusnasstrom (approved):
Looks great Haoyu, thank you!

lauri-codes (approved):

PR #250, opened 2026-03-13
Add human in the loop and asset ref docs

blueraft:


lauri-codes:
This should target v1.4.3: don't merge yet

lauri-codes (approved):

PR #251, opened 2026-03-16
add docs about scopes and tokens

DanielYang59:


docs/explanation/auth.md
```diff
@@ -0,0 +1,123 @@
...
+    L --> M
+
+    M -- Yes --> N[Request allowed]
+    M -- No --> O[HTTP 403 Forbidden<br/>Missing scopes]
+
+    class D,E,H,J,L config
+    class N success
+
+    classDef config fill:#fff5f5,stroke:#d73a49,stroke-width:2px;
+    classDef success fill:#e6ffed,stroke:#2ea44f,stroke-width:2px;
```
lauri-codes, line 49: This is very helpful, and mermaid is the correct way to do this. Can the text size be increased so that the text would be readable without zoom?

docs/explanation/auth.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+## Access tokens
+
+NOMAD supports several types of tokens for authenticating API requests:
```
lauri-codes: Modify to: "NOMAD supports several types of access tokens for authenticating and authorizing API requests:"

docs/explanation/auth.md
```diff
@@ -0,0 +1,123 @@
...
+- Keycloak access tokens – obtained through the login flow used by the web UI or API clients
+- Personal Access Tokens (PATs) – long-lived tokens created by users for scripts and automation
+- Upload tokens (*legacy, deprecated* — will be removed in a future release)
+- Simple (app/signature) tokens (*legacy, deprecated* — will be removed in a future release)
+
+### Keycloak access tokens
+
+Keycloak access tokens authenticate a user through the NOMAD identity provider `keycloak`.
+
+Currently, the backend treats authenticated users as having the full set of user scopes.
```
lauri-codes: You could tell some more details here:

- Retrieved from Keycloak using OpenID Connect (OIDC) 
- Only have a very limited timespan (X minutes by default, can be refreshed), not suitable for long-term access
- Are used when user is interacting with the server using the GUI
- These tokens have access to all regular user scopes, no possibillity to restrict access

docs/explanation/auth.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Simple tokens
+
+Simple tokens are legacy, deprecated tokens.
+They grant a broad set of user permissions, effectively corresponding to full access
+except for token-management scopes.
+
+!!! warning "Deprecated"
+    Simple tokens are deprecated and will be removed in a future release.
+    They should not be used for new integrations. Use **Personal Access Tokens (PATs)** instead.
```
lauri-codes: You could put these e.g. in an info admonition to make this section less prominent: these will be deprecated.

docs/explanation/auth.md
```diff
@@ -0,0 +1,123 @@
...
+```text
+uploads:read
+uploads:write
+```
+
+When an API endpoint is called, NOMAD checks whether the request has the scopes required by that endpoint.
+If the required scopes are missing, the request is rejected.
+
+For example, if an endpoint requires `uploads:write` and the token only has `uploads:read`,
+the request will be rejected.
```
lauri-codes: We need to have a table of all possible scopes. Ideally this should be automatically built from our codebase, but I'm not sure how easy this will be. Once we have this, we can in the GUI, in the admin access section link to this table.

docs/explanation/auth.md
```diff
@@ -0,0 +1,123 @@
...
+
+!!! warning "Deprecated"
+    Simple tokens are deprecated and will be removed in a future release.
+    They should not be used for new integrations. Use **Personal Access Tokens (PATs)** instead.
+
+## Authorization via scopes
+
+Authorization in NOMAD is scope-based.
+After authentication, NOMAD determines the effective **scopes** that apply to the request.
+Scopes define which API operations a user or token is allowed to perform.
```
lauri-codes: Authorization to **use specific API features is scope-based**. There are also certainly parts of authorization that are based on the user, user group, or being the admin user.

lauri-codes (changes requested):
Looks very nice, this is a good addition and I think docs/explanation is the right place.

A few comments, have a look.

docs/explanation/auth.md
DanielYang59: I tried here and maybe I would need to add a macro because currently we don't have one for enum https://github.com/FAIRmat-NFDI/nomad-docs/pull/241#discussion_r2929961968

docs/explanation/auth.md
```diff
@@ -0,0 +1,138 @@
...
+
+```mermaid
+%%{init:{
+  'themeVariables': {
+    'fontSize': '36px'
+  }
+}}%%
+flowchart TD
+    A[Request reaches Oasis] --> B{Network access allowed by deployment?}
+    B -- No --> X[Access blocked outside NOMAD]
```
DanielYang59, line 21: looks like with larger font, current line wrapping a bit aggressive

DanielYang59:
@hampusnasstrom if you have time maybe you could have a look here as well?

docs/explanation/auth.md
```diff
@@ -0,0 +1,142 @@
...
+well suited for long-running scripts or integrations where repeatedly acquiring
+new tokens would be inconvenient.
+
+PAT scopes must always be explicitly specified. Wildcard expressions such as
+`*:read` are supported only in configuration and are **not allowed in tokens**.
+
+For programmatic API usage, **Personal Access Tokens are generally the recommended
+authentication method**.
+
+### Upload tokens
```
DanielYang59: TODO: I forgot to put the entire section under deprecation tag

lauri-codes (approved):

PR #255, opened 2026-03-19
Auto light dark mode

DanielYang59:
- docs page support light/dark/auto mode
- pin mkdocs < 2

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -115,8 +115,24 @@ nav:
 theme:
   name: material
   palette:
-    primary: "#2A4CDF"
-    accent: "#008A67"
+    - media: "(prefers-color-scheme)"
+      toggle:
+        icon: material/brightness-auto
+        name: Switch to light mode
+    - media: "(prefers-color-scheme: light)"
+      scheme: default
+      primary: custom
+      accent: custom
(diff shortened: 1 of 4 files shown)


JFRudzinski (approved):
Feature is great 👍 

Not sure about having dark mode the default, for me I prefer it, but it's more of a brand or ui/ux thing, please check with @lauri-codes and @g-berfinguener

mkdocs.yml
```diff
@@ -115,8 +115,24 @@ nav:
 theme:
   name: material
   palette:
-    primary: "#2A4CDF"
-    accent: "#008A67"
+    - media: "(prefers-color-scheme)"
```
DanielYang59, line 118: I believe auto mode is default here, did you happen to be using dark mode in your browser?

JFRudzinski: ah ok, nice! didn't realize, yes I am sure I do ... then I think it is fine to merge in that case

DanielYang59: the behaviour is now changed in https://github.com/FAIRmat-NFDI/nomad-docs/pull/268 now to use light mode by default (to be consistent with the new GUI)

PR #259, opened 2026-03-24
fix northtoolentrypoint anchor

DanielYang59:
this should be lower case:
```
INFO    -  Doc file 'howto/plugins/types/north_tools.md' contains a link
           '../../../reference/plugins.md#NORTHToolEntryPoint', but the doc 'reference/plugins.md' does not contain
           an anchor '#NORTHToolEntryPoint'.
```


lukaspie (approved):
LGTM, thanks!

PR #260, opened 2026-03-25
Add how-to page for using MADAS for data analytics

kubanmar:
Adding a page that introduces MADAS for downloading data from NOMAD. It features a short introduction with two figures and links to the MADAS documentation and tutorials.


docs/howto/manage/program/MADAS.md
```diff
@@ -0,0 +1,19 @@
+# Data analytics with `MADAS`
+
+For more advanced machine learnining and data analytics tasks, it may be beneficial to download and cache data from NOMAD locally. This functionality is supported by the MAterials DAta Similarity (`MADAS`) framework.
+
+<p align="center" width="100%">
+    <img src="./images/madas_overview.png" width="50%" alt="Overview of MADAS"></img>
+</p>
+
+Using `MADAS`, data can be downloaded, transformed, and analyzed, making it a comprehensive framework for working with data from NOMAD. To simplify the analyis, starting from the rather verbose schema of within NOMAD, it allows to define functions that map selected quantities to a new schema. After downloading the data, it can be stored in a local database, ensuring reproducibility of results within the analysis pipeline. The stored data can be used to derive fingerprints, which allow to represent the data for different data analytics and machine learning tasks.
```
JFRudzinski: there is a typo in "analysis" -- can you do a full spell check please

JFRudzinski (approved):
For me this is good as is. There are some questions in my head about long-term placement of external tools vs. native NOMAD functionality -- but I think this should not hold back the initial publishing of this page. 

However, I would like to ask Laurie for a sign-off as well, please wait for him before merging

lauri-codes (commented):
I think this will make things more confusing for users.

MADAS is not an "official" FAIRmat product (does not live in the FAIRmat-NFDI organization and is not the default query tool we endorse), and thus it does not fit into our main docs. If we include it here, users will be confused even more by the abumdance of different query helper tools (we already have ArchiveQuery + workflow tool helpers).

A better approch to handling this would be to have a doc section that lists third-party NOMAD tools. There we can definitely have a a link + description of MADAS.

kubanmar:
That sounds like a valid concern. I can move the description to a new section. @JFRudzinski , where could that live?

For completeness, I would like to point out that MADAS is not a query tool, but a more complete framework for analyzing data.

JFRudzinski:
I agree, and I am starting to make some plans for docs refactoring ... but this will take some time still. We need to decide if we want to throw something up in the mean time. We could just put it under the Computational section for now. @kubanmar let's just discuss in person next week please

PR #262, opened 2026-03-26
Remove some unused assets, relocate some non-global items from `assets`

DanielYang59:


docs/howto/oasis/configure.md
```diff
@@ -550,7 +550,7 @@ installation above. There are just a three changes.
 - The `nomad.yaml` has modifications to tell Oasis to use your and not the official NOMAD keycloak.
 
 You can start with the regular installation above and manually adopt the config or
-download the already updated configuration files: [nomad-oasis-with-keycloak.zip](../../assets/nomad-oasis-with-keycloak.zip).
+download the already updated configuration files: [nomad-oasis-with-keycloak.zip](./ops/nomad-oasis-with-keycloak.zip).
```
JFRudzinski: `ops/` -> `data/` ? see https://github.com/FAIRmat-NFDI/nomad-docs/blob/develop/docs/writing_guide.md#repo-organization

Or do you feel these assets are fundamentally different somehow?

DanielYang59: Hi I don't have strong opinion either way (as long as it's out of the global assets folder), I used ops because it's already there (plus it's an example config file)

JFRudzinski: but do you know why it is there? What is the other content? If it is all just example data files, then I would just change to data/

I also don't care about the exact naming as a whole, but I do care about the standardization because it is easier to understand and maintain -- so if no one is naming something for a particular reason then we should keep with our standard

DanielYang59: > but do you know why it is there? What is the other content? If it is all just example data files, then I would just change to data/

I think the rest is example config files ([used in `howto/oasis/configure.md`](https://github.com/FAIRmat-NFDI/nomad-docs/blob/develop/docs/howto/oasis/configure.md?plain=1#L53)), I guess they must be copied from other part of our codebase where `ops` is more standard there (for example in [`nomad-fair`](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/tree/develop/ops?ref_type=heads)).

I have renamed them to `data` to follow the docs repo standard here

JFRudzinski (approved):

PR #267, opened 2026-03-30
add docs for Auth in plugins

### LINKED ISSUE #245: auth for plugin entry point
https://nomad-lab.eu/prod/v1/docs/howto/plugins/types/apis.html

cc @yaruwangway

DanielYang59:
close https://github.com/FAIRmat-NFDI/nomad-docs/issues/245

https://discord.com/channels/1201445470485106719/1481239096956293120

### DIFF
--- a/docs/howto/plugins/types/apis.md
+++ b/docs/howto/plugins/types/apis.md
@@ -122,3 +122,49 @@ curl localhost:8000/nomad-oasis/myapi/static/static_page.html
     distributed e.g. in PyPI, you will need to explicitly include them in the
     `MANIFEST.in` file of your Python package. See more information in [the setuptools
     guide](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html#controlling-files-in-the-distribution){:target="_blank" rel="noopener"}.
+
+## Protecting the API with auth
+
+If your API should only be accessible to authenticated requests that satisfy the
+configured access policy (deployment whitelist and scopes), you can attach an auth dependency
+directly to the `FastAPI` app.
+This ensures that every request:
+
+- Is made by an authenticated user (authentication)
+- Has the required permissions via [scopes](../../../explanation/auth.md#authorization-via-scopes) (authorization)
+
+The dependency `get_current_user(...)` enforces both.
+
+Use app-level dependencies when all endpoints share the same access requirements,
+and route-level dependencies when different endpoints require different scopes.
+
+For example, the following app requires the "uploads:read" scope for every request:
+
+```python hl_lines="6"
+from nomad.app.v1.routers.auth import get_current_user
+from nomad.auth.scopes import Scope
+
+app = FastAPI(
+    ...,
+    dependencies=[Depends(get_current_user([Scope.UPLOADS_READ], allow_anonymous=False))],
+)
+```
+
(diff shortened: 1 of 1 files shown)


DanielYang59:
@yaruwangway @hampusnasstrom I added a draft here, can you have a look and tell me if there's anything unclear?

docs/howto/plugins/types/apis.md
```diff
@@ -122,3 +122,48 @@ curl localhost:8000/nomad-oasis/myapi/static/static_page.html
     distributed e.g. in PyPI, you will need to explicitly include them in the
     `MANIFEST.in` file of your Python package. See more information in [the setuptools
     guide](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html#controlling-files-in-the-distribution){:target="_blank" rel="noopener"}.
+
+## Protecting the API with auth
+
+If your API should only be accessible to certain users, you can attach an
```
yaruwangway: this "certain users" means authenticated or not, right ? it's not possible to make i make a userA access to the app but userB not.

DanielYang59: Maybe it's a bit unclear here, I would clarify this.

Currently the whitelist authorization could only be setup globally to the entire deployment (via `nomad.yaml`), so if that deployment has setup whitelist, `get_current_user` would enforce it, but currently it's not possible to inject different whitelists directly via the dependency

docs/howto/plugins/types/apis.md
```diff
@@ -122,3 +122,49 @@ curl localhost:8000/nomad-oasis/myapi/static/static_page.html
...
+
+## Protecting the API with auth
+
+If your API should only be accessible to authenticated requests that satisfy the
+configured access policy (deployment whitelist and scopes), you can attach an auth dependency
+directly to the `FastAPI` app.
+This ensures that every request:
+
+- is made by an authenticated user (authentication)
+- has the required permissions via [scopes](../../../explanation/auth.md#authorization-via-scopes) (authorization)
```
lauri-codes: Capitalize first letter of each item

docs/howto/plugins/types/apis.md
```diff
@@ -122,3 +122,49 @@ curl localhost:8000/nomad-oasis/myapi/static/static_page.html
...
+For example, the following app requires the "uploads:read" scope for every request:
+
+```python hl_lines="6"
+from nomad.app.v1.routers.auth import get_current_user
+from nomad.auth.scopes import Scope
+
+app = FastAPI(
+    ...,
+    dependencies=[Depends(get_current_user([Scope.UPLOADS_READ], allow_anonymous=False))],
+)
```
lauri-codes, line 150: Ideally these code snippets would be in separate files that are actually linted and at least run once during the docs tests: this would ensure that we don't show non-functional snippets.

If you think it is overkill, we can also do the inline code as well.

DanielYang59: yes I agree we should try to lint/format or better actually test them, but I guess it's a global issue to the entire docs repo, I would open an issue to track this  (maybe not in the PR?)
- https://github.com/FAIRmat-NFDI/nomad-docs/issues/270

lauri-codes (approved):

PR #290, opened 2026-05-21
Update auth guide

### LINKED ISSUE #285: Update "programmatic" guide for auth
- [ ] some guide around programatic are outdated after recent PAT changes
- [ ] include guide on auth with PAT via other tools like `curl` (and `curlrc` with URL matching), and usage of PAT via env var

DanielYang59:
close https://github.com/FAIRmat-NFDI/nomad-docs/issues/285

### DIFF
--- a/docs/howto/manage/program/auth.md
+++ b/docs/howto/manage/program/auth.md
@@ -52,7 +52,10 @@ raw_token = pat["raw_token"]
 
 ### Use a PAT
 
-Once created, we recommend [storing it in an environment variable](https://developer.android.com/tools/variables){:target="_blank" rel="noopener"}, for example in macOS/Linux:
+Once created, we recommend [storing it in an environment variable](https://en.wikipedia.org/wiki/Environment_variable){:target="_blank" rel="noopener"}, for example in macOS/Linux:
+
+This helps avoid hardcoding secrets in scripts, which can be accidentally leaked
+through version control, logs, or shared files.
 
 ```bash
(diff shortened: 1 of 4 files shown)


docs/howto/manage/program/auth.md
```diff
@@ -50,14 +50,16 @@ pat = response.json()
 raw_token = pat["raw_token"]
 ```
 
-### Use a PAT in Python
+### Use a PAT
 
-Once created, we recommend storing it in an environment variable:
+Once created, we recommend [storing it in an environment variable](https://developer.android.com/tools/variables){:target="_blank" rel="noopener"}, for example in macOS/Linux:
```
lauri-codes: The link is a bit odd: why Android docs...? We should use some more generic doc page about env variables, e.g. https://en.wikipedia.org/wiki/Environment_variable

Also, we should hare epxlicitly say **why** we recommend this: this way you don't accidentally store the sensitive information in files that may end up in version controlled files etc.

DanielYang59: > The link is a bit odd: why Android docs...? We should use some more generic doc page about env variables, e.g. https://en.wikipedia.org/wiki/Environment_variable


I tried to search for a more practical guide (without too much low level details) from a trusted source (and it doesn't have too much Android-specific things and jumped directly into how to use it in various platforms)... Looks like I could link directly to the "syntax" section in wikipedia otherwise people might be confused with all those instruction/design details

lauri-codes (approved):

PR #292, opened 2026-05-26
Add dashboard plugin entry points documentation and update navigation

### LINKED ISSUE #291: Add docs for Dashboard plugin entry point
We should add documentation for the new Dashboard plugin entry point.

This first iteration of the docs should cover just the basics:

- What are dashboards? What can they be used for? How do they differ from APIs?
- Difference between hosting the dashboard via FastAPI vs hosting it externally
- Launch modes
- Authentication: how it works with FastAPI (cookies sent automatically, just use the FastAPI dependency to capture them). Mention that we don't yet have a robust solution for externally hosted dashboards.
- Styling: What can a user do to make the Dashboard fit with the NOMAD styles? Can the dashboard somehow know the selected theme...? Can the dashboards read any of the MUI CSS variables? Where to find our theme colors: link to the theme file we have in the new GUI (we ignore old GUI, as Dashboards are only availble in new GUI)

amirgolp:
Closes #291

### DIFF
--- a/docs/howto/plugins/types/dashboards.md
+++ b/docs/howto/plugins/types/dashboards.md
@@ -1,0 +1,494 @@
+# How to write a dashboard
+
+## What is a dashboard?
+
+A **dashboard** is a custom web interface that a NOMAD plugin contributes
+to the NOMAD GUI. Users discover installed dashboards on the *Dashboards*
+page of the GUI and can open each one either **embedded** (inside an iframe
+on a dedicated viewer page in the GUI) or in a **new browser tab**.
+
+Dashboards can be used for a wide variety of purposes:
+
+- A small interactive search or filtering UI tailored to a specific
+  domain or community.
+- A custom data-viewer or analysis tool (e.g. a Jupyter-like notebook,
+  a plotting tool, a chemistry-specific 3D viewer).
+- A landing page or guided workflow for a particular user group.
+- A bridge to an existing internal tool that should be reachable from
+  within NOMAD without forcing users to remember a separate URL.
+
+### How do dashboards differ from APIs?
+
+[API plugin entry points](./apis.md) and dashboards are technically
+similar (both can mount a FastAPI app inside NOMAD), but their audience
+is different:
+
+| | **APIs** | **Dashboards** |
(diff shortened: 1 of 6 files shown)


docs/howto/plugins/types/dashboards.md
```diff
@@ -0,0 +1,494 @@
...
+### How do dashboards differ from APIs?
+
+[API plugin entry points](./apis.md) and dashboards are technically
+similar (both can mount a FastAPI app inside NOMAD), but their audience
+is different:
+
+| | **APIs** | **Dashboards** |
+| --- | --- | --- |
+| Audience | Other programs (clients, scripts, integrations) | End users via a browser |
+| Surfaces in the GUI? | No (just an extra `/myapi/` mount) | Yes — listed on the *Dashboards* page, with embedded + new-tab launchers, icons, descriptions |
```
lauri-codes: Not true: also APIs have a dedicated page where they can be browsed, and we provide links e.g. to the dashboards.

docs/howto/plugins/types/dashboards.md
```diff
@@ -0,0 +1,494 @@
...
+| Typical content | JSON endpoints | HTML + JS, optionally a full SPA |
+| Can be external? | No — must return a FastAPI app | Yes — can point at an external URL |
```
lauri-codes: We will add the support for external APIs as well, so don't include this line.

lauri-codes: Probably a table no longer makes sense if we then have only two rows...?

lauri-codes (commented):
Overall very good! I think this covers most of the questions that users will have.

docs/howto/plugins/types/dashboards.md
amirgolp: make sense, I have removed the table and replaced with bulletpoints

docs/howto/plugins/types/dashboards.md
amirgolp: updated

lauri-codes (approved):

PR #303, opened 2026-06-10
Improve visibility of get_plugin_entry_point function

ndaelman-hu:
Adds a new section 'Accessing plugin configuration at runtime' to the plugin development guide, documenting how to use `get_plugin_entry_point()` to retrieve plugin configuration, the lazy-loading pattern, and when to use this function. This addresses the low discoverability of this important function which was previously only mentioned briefly in scattered locations.

### DIFF
--- a/docs/howto/plugins/plugins.md
+++ b/docs/howto/plugins/plugins.md
@@ -65,7 +65,7 @@ plugins:
 
 ## Accessing plugin configuration at runtime
 
-When developing plugins, you can access both the plugin's configuration (metadata and parameters) and the actual plugin implementation (the resource). NOMAD uses a [lazy-loading pattern](../../explanation/plugin_system.md#plugin-resource) where these are kept separate: `get_plugin_entry_point()` returns the entry point configuration, while `.load()` returns the actual plugin resource.
+When developing plugins, you can access both the plugin's configuration (metadata and parameters) and the actual plugin implementation (the resource). NOMAD uses a [lazy-loading pattern](../../explanation/plugin_system.md#plugin-resource) where these are kept separate: `get_plugin_entry_point()` returns the entry point configuration, while `EntryPoint.load()` returns the actual plugin resource.
 
 ### Configuring plugin parameters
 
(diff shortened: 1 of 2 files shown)


ndaelman-hu:
@lauri-codes idk who else designed the plugin system, so feel free to tag

docs/howto/plugins/plugins.md
```diff
@@ -63,6 +63,73 @@ plugins:
...
+
+```python
+from nomad.config import config
+
+entry_point = config.get_plugin_entry_point('nomad_example.parsers:myparser')
+parser = entry_point.load()
+
+# Use the parser instance
+parser.parse('path/to/file.out', archive, logger)
+```
```
lauri-codes: I would maybe not include this section: users should almost never load the resources manually: NOMAD handles this automatically, and I think users can do more harm by loading the resources like this (might break lazy-loading, might cause resource duplication due to caching, etc.).

docs/howto/plugins/plugins.md
```diff
@@ -63,6 +63,73 @@ plugins:
...
+
+# Get your plugin's entry point configuration
+entry_point = config.get_plugin_entry_point('nomad_example.parsers:myparser')
+
+# Access configuration parameters
+print(f'Parameter value: {entry_point.parameter}')  # Output: Parameter value: custom_value
+print(f'Another setting: {entry_point.another_setting}')  # Output: Another setting: 42
+
+# Access metadata
+print(f'Plugin name: {entry_point.name}')  # Output: Plugin name: nomad_example.parsers:myparser
```
lauri-codes: I would not make a separation here between parameter and metadata: just makes it more confusing in this case. Everythign in a parser except the ID is re-configurable.

Thus I suggest removing lines 100 and 101, and the modifying the text on line 87 e.g. to: "...giving you access to custom parameters and other relevant configuration"

lauri-codes (commented):
Few small tweaks, looks good!

docs/howto/plugins/plugins.md
ndaelman-hu: Okay, I just added it since it comes as the natural question after "here's how you load  the configs".
There is a disclaimer already, but I could either increase its visibility with a warning admonition, or remove this section altogether.

docs/howto/plugins/plugins.md
ndaelman-hu: > Everythign in a parser except the ID is re-configurable.

Okay, that is actually pretty interesting. Only via the `EntryPoint` class or are all configurations modifiable via the `nomad.yaml` too?

lauri-codes: Everything is modifiable via `nomad.yaml`

docs/howto/plugins/plugins.md
lauri-codes: For now, remove altogether.

docs/howto/plugins/plugins.md
ndaelman-hu: Suggestion applied. Also removed sentence below line 87, as it reiterated the use case.

docs/howto/plugins/plugins.md
ndaelman-hu: Applied

ndaelman-hu:
> Few small tweaks, looks good!

All suggestions applied. Ready to merge.

lauri-codes (approved):

PR #308, opened 2026-07-07
Stabilize docs validation by building MkDocs output during tests and …

JFRudzinski:
This PR restores the documentation validation baseline.

It fixes tests/test_docs.py so the docs-serving test builds the current MkDocs site instead of relying on prepackaged static docs. 

It also corrects the README test command from uv run --extra dev pytest to uv run --group dev pytest -q, since dev is configured as a dependency group in this repo, not an optional extra.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,9 @@ uv run mkdocs build --strict
 uv run --group dev pytest -q
 ```
 
-Temporary note for the current refactoring project: when you touch deployment-specific content, add an HTML comment near the top of the page to label scope without changing rendered output, for example `<!-- doc-scope: deployment-specific; target=oasis -->`.
+### Temporary refactoring note
+
+For the current refactoring project, when you touch deployment-specific content, add an HTML comment near the top of the page to label scope without changing rendered output, for example `<!-- doc-scope: deployment-specific; target=oasis -->`.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -81,10 +81,6 @@ uv run mkdocs build --strict
 uv run --group dev pytest -q
 ```
 
-### Temporary refactoring note
-
-For the current refactoring project, when you touch deployment-specific content, add an HTML comment near the top of the page to label scope without changing rendered output, for example `<!-- doc-scope: deployment-specific; target=oasis -->`.
-
 ---
 
 ## Automated Tests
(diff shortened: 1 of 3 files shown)


JFRudzinski:
@DanielYang59 I'm not so sure about this approach, but something was wrong with the old test setup when I try to run locally ... let me know if you have other suggestions

README.md
```diff
@@ -77,7 +77,8 @@ This will install all requirements in a virtual environment and start the local
 ### How to run the tests
 
 ```bash
-uv run --extra dev pytest
+uv run mkdocs build --strict
+uv run --group dev pytest -q
```
DanielYang59, line 81: the `dev` group doesn't need to be included manually, it's [included by default](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups):
> By default, uv includes the dev dependency group in the environment (e.g., during uv run or uv sync).

tests/test_docs.py
```diff
@@ -1,11 +1,44 @@
...
+from nomad.app.static import app as static_app
 from nomad.config import config
 
 
 @pytest.fixture(scope="session")
-def client():
+def built_docs_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
+    repo_root = Path(__file__).resolve().parent.parent
+    site_dir = tmp_path_factory.mktemp("site")
+    subprocess.run(
```
DanielYang59, line 16: > something was wrong with the old test setup when I try to run locally

I guess this might be somewhat "expected" as the `nomad-lab` package doesn't pack up docs, therefore in CI we have: https://github.com/FAIRmat-NFDI/nomad-docs/blob/484de85ce81b7a7bc69a30b35cf3cdff783c57ea/.github/workflows/test-build-lint-preview.yml#L59-L74

And if we add this fixture here, then docs would be built twice (though we might remove the other build step in CI)...


But I guess in most cases contributors doesn't really need to run `test_docs` locally as it mostly test how docs is served and maybe just let CI run it? maybe we could add a conditional skip mark like `@pytest.mark.skipif(not os.getenv("CI"), reason="only runs in CI")`?

Any comment? @blueraft

JFRudzinski: if we think it's unnecessary to run the tests locally, we could just close this ... this is why I said I was unsure. I was just doing a full audit of the repo and the agent of course wanted to test things locally

JFRudzinski:
closed as unnecessary for contributors to run local testing suite

PR #319, opened 2026-07-16
Plugin navigation revived

JFRudzinski:
revived by main merge into #186


JFRudzinski:
close in favor of reverting back to original PR #186

PR #323, opened 2026-07-20
Fix nomad_url() emitting localhost URLs in published docs

siamakn:
Set the instance config (`NOMAD_SERVICES_*` env) in the deploy build so the macro renders `https://nomad-lab.eu/prod/v1/api/...`. The macro is unchanged, so local/Oasis builds still produce their own URL. This fixes a
incorrectly generated link on the API page, plus adds a CI guard that fails if a localhost URL leaks.

This changes the build config, not the already-published HTML mkdocs page, so the live docs update only on the next deploy build.

### DIFF
--- a/.github/workflows/test-build-lint-preview.yml
+++ b/.github/workflows/test-build-lint-preview.yml
@@ -66,18 +66,14 @@ jobs:
           python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Check mkdocs build
-        run: |
-          uv run \
-            --with 'nomad-lab[infrastructure] @ https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR.git' \
-            mkdocs build --strict
-
-      - name: Check production build has no localhost URLs
         # The `nomad_url()` macro renders relative to the configured instance.
-        # The production deploy sets these env vars so docs point at
-        # https://nomad-lab.eu/prod/v1/api/... (see deploy-docs.yml). Build with
-        # the same config and fail if the Oasis default (localhost/nomad-oasis)
-        # leaks into the published output. `.linkspector` intentionally ignores
-        # localhost, so it cannot catch this class of regression.
+        # Set the same env vars as the production deploy (see deploy-docs.yml) so
+        # `site/` points at https://nomad-lab.eu/prod/v1/api/... — this is the
+        # directory that is copied into the nomad package and published as the PR
+        # preview below, so it must be built with the production config. Then fail
+        # if the Oasis default (localhost/nomad-oasis) leaks into the output.
+        # `.linkspector` intentionally ignores localhost, so it cannot catch this
+        # class of regression.
         env:
           NOMAD_SERVICES_API_HOST: nomad-lab.eu
           NOMAD_SERVICES_API_PORT: "443"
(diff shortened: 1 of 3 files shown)


.github/workflows/test-build-lint-preview.yml
```diff
@@ -71,6 +71,31 @@ jobs:
             --with 'nomad-lab[infrastructure] @ https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR.git' \
             mkdocs build --strict
 
+      - name: Check production build has no localhost URLs
```
DanielYang59, line 65: Perhaps it's unnecessary to have a dedicated step just for checking URLs, this should be doable inside the above "Check mkdocs build" step?

.github/workflows/test-build-lint-preview.yml
```diff
@@ -71,6 +71,31 @@ jobs:
...
+        # The `nomad_url()` macro renders relative to the configured instance.
+        # The production deploy sets these env vars so docs point at
+        # https://nomad-lab.eu/prod/v1/api/... (see deploy-docs.yml). Build with
+        # the same config and fail if the Oasis default (localhost/nomad-oasis)
+        # leaks into the published output. `.linkspector` intentionally ignores
+        # localhost, so it cannot catch this class of regression.
+        env:
```
DanielYang59, line 72: These env var are only applied to the "Check production build has no localhost URLs" step, and "Check mkdocs build" didn't have them so the previous site (using the `site` directory from the previous "Check mkdocs build" step) still has the broken URLs

DanielYang59 (changes requested):

.github/workflows/test-build-lint-preview.yml
```diff
@@ -71,6 +62,31 @@ jobs:
...
```
DanielYang59, line 72: perhaps we could define repo level `vars` to avoid duplicating them?
https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#defining-configuration-variables-for-multiple-workflows

blueraft:
This wouldn't change the URL on the actual production website since the docs there is built in the nomad-distro pipelines. 

There's also some discussion on hosting the docs on the main nomad website, I'm not sure if there's a date set for that. @JFRudzinski do you know?

JFRudzinski:
> This wouldn't change the URL on the actual production website since the docs there is built in the nomad-distro pipelines.
> 
> There's also some discussion on hosting the docs on the main nomad website, I'm not sure if there's a date set for that. @JFRudzinski do you know?

There is no definite date, but my target would be to have the hompage refactoring with a good draft by end of Aug and basically complete by end of Sep ... I'm not sure exactly when the docs deployment would come in, but also by end of Sep

blueraft:
We're now deploying docs from this repo, so this can be rebased and merged once @DanielYang59 approves the changes

siamakn:
Thanks @DanielYang59
So I moved the env vars onto Check mkdocs build (the site/ we actually deploy). Checked locally: with the vars it renders nomad-lab.eu/prod/v1/api/...; without them the guard still catches localhost:8000/nomad-oasis/api.
Also the set the check step in mkdocs build, dropped the extra site-prod build.
 I differed for now the repo-level vars as a follow-up  task since it needs the vars set in repo Settings.

DanielYang59:
@siamakn thanks! Can you rebase (there was some change on the deployment workflow) and I would review.

.github/workflows/test-build-lint-preview.yml
```diff
@@ -66,10 +66,30 @@ jobs:
...
+        # `site/` points at https://nomad-lab.eu/prod/v1/api/... — this is the
+        # directory that is copied into the nomad package and published as the PR
+        # preview below, so it must be built with the production config. Then fail
+        # if the Oasis default (localhost/nomad-oasis) leaks into the output.
+        # `.linkspector` intentionally ignores localhost, so it cannot catch this
+        # class of regression.
+        env:
+          NOMAD_SERVICES_API_HOST: nomad-lab.eu
+          NOMAD_SERVICES_API_PORT: "443"
+          NOMAD_SERVICES_API_BASE_PATH: /prod/v1
```
blueraft, line 80: Add a todo here to change this to '/' once the new version is out

blueraft: Same in deploy-docs.yml

PR #326, opened 2026-07-24
repointed the account creation links to correct location in How-to s

### LINKED ISSUE #325: Update the links in the tutorials pages to the create an account
[As PR 137 ](https://github.com/FAIRmat-NFDI/nomad-docs/pull/309) move the create an account information to the how-to guides. 

This change needs to be reflected throughout the tutorial pages, which currently refer to the overview page.

siamakn:
repointed the account creation links to corresponding location in the how-to s

### DIFF
--- a/docs/tutorial/eln/built_in_templates.md
+++ b/docs/tutorial/eln/built_in_templates.md
@@ -28,7 +28,7 @@ Before starting, make sure you have:
 1. **NOMAD user account**  
    Creating and editing ELN entries requires a NOMAD user account.
    You can create an account by following the steps described in the
-   [overview page](../overview.md#create-a-nomad-user-account).
+   [account creation guide](../../howto/manage/gui/account.md#create-a-nomad-account).
 
 2. **Basic understanding of uploads and entries**  
    Familiarity with uploads, entries, and how they relate to each other can be helpful. These concepts are introduced in the section [key elements in NOMAD](../upload_publish.md#the-key-elements-in-nomad) and will be reinforced throughout the tutorial.
--- a/docs/tutorial/eln/built_in_templates.md
+++ b/docs/tutorial/eln/built_in_templates.md
@@ -28,7 +28,7 @@ Before starting, make sure you have:
 1. **NOMAD user account**  
    Creating and editing ELN entries requires a NOMAD user account.
    You can create an account by following the steps described in the
-   [account creation guide](../../howto/manage/gui/account.md#create-a-nomad-account).
+   [How-to guides > ... > Create a NOMAD account](../../howto/manage/gui/account.md#create-a-nomad-account).
 
 2. **Basic understanding of uploads and entries**  
    Familiarity with uploads, entries, and how they relate to each other can be helpful. These concepts are introduced in the section [key elements in NOMAD](../upload_publish.md#the-key-elements-in-nomad) and will be reinforced throughout the tutorial.
--- a/docs/tutorial/eln/custom_eln_yaml.md
+++ b/docs/tutorial/eln/custom_eln_yaml.md
@@ -26,7 +26,7 @@ Before starting, make sure you have:
 1. **NOMAD user account**  
    Creating and editing ELN entries requires a NOMAD user account.
    You can create an account by following the steps described in the
-   [overview page](../overview.md#create-a-nomad-user-account).
+   [account creation guide](../../howto/manage/gui/account.md#create-a-nomad-account).
 
 2. **Basic understanding of uploads and entries**  
    Familiarity with uploads and entries, and with how they relate to each other can be helpful. These concepts are introduced in the section [key elements in NOMAD](../upload_publish.md#the-key-elements-in-nomad).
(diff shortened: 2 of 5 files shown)


docs/tutorial/eln/built_in_templates.md
```diff
@@ -28,7 +28,7 @@ Before starting, make sure you have:
 1. **NOMAD user account**  
    Creating and editing ELN entries requires a NOMAD user account.
    You can create an account by following the steps described in the
-   [overview page](../overview.md#create-a-nomad-user-account).
+   [account creation guide](../../howto/manage/gui/account.md#create-a-nomad-account).
```
JFRudzinski, line 31: Let's always use the name of the actual page or section for consistency, e.g., in this case "...described in [Create a NOMAD account]" ... see also "Internal links: standardized naming." section of the [writing guide](https://github.com/FAIRmat-NFDI/nomad-docs/blob/main/docs/writing_guide.md) for further standards of link texts using ">" notation

JFRudzinski (changes requested):
I took over the review here since Ahmed is on vacation, we shouldn't wait to get this small PR through. Can you address my comment and then tag me for review. We can then get it through quite quickly

We need to systematically address the standard link names, but let's at least try to harmonize anything new that we add

siamakn:
Thanks @JFRudzinski for the review.
I was on vacations, now  back and fixed them to How-to guides > ... > Create a NOMAD account" in 5 locations.

docs/tutorial/eln/built_in_templates.md
```diff
@@ -28,7 +28,7 @@ Before starting, make sure you have:
 1. **NOMAD user account**  
    Creating and editing ELN entries requires a NOMAD user account.
    You can create an account by following the steps described in the
-   [overview page](../overview.md#create-a-nomad-user-account).
+   [How-to guides > ... > Create a NOMAD account](../../howto/manage/gui/account.md#create-a-nomad-account).
```
DanielYang59, line 31: Is this correct (or shortened in the right way because it's quite deeply nested)? it [looks like](https://docs.nomad-lab.eu/pr-preview/pr-326/howto/manage/gui/account.html):
`How-to guides > Manage and find data > Use the GUI > Create an account and sign in`

I assume it could be hard for new users to guess what is `...` because each level has multiple entries (I had to go to that URL to find it...)...

image

JFRudzinski: Well, this is what I proposed in the writing guide, but we can of course adjust it if there is a better way. I wasn't thinking that it is supposed to tell the user the exact path via the navigation tab, rather it indicates to them that they are (perhaps) moving to a new main section of the docs (How-to vs. Reference vs. ...). That being said I do see that it you are referencing a page section, you might want to always state explicitly the page

I was just trying to avoid having these very long paths, but also having some sort of standard. What would your suggestion be?

DanielYang59: > That being said I do see that it you are referencing a page section, you might want to always state explicitly the page
> I was just trying to avoid having these very long paths, but also having some sort of standard.

I guess from our (developer) prospective it might looks cleaner to have a shorter path, and I'm not sure if users (readers of docs) would benefit from this... I assume users might be happy as long as the tutorial easily guides them to the sections they need to find, in this case `How-to guides > ... > Create a NOMAD account` might fail to achive this because when they click on `How-to guides` it might not be very clear which section `Create a NOMAD account` is in...


or perhaps if searching is preferred we could link directly to the final section name... but I really don't know, and I personally wouldn't mind a longer path...

image

JFRudzinski: I guess the answer comes down to what we are trying to achieve here:

- standardization: this is the primary motivation. Before all sorts of different short names were being used in the link text, many not using the page or section name at all.
- user experience: I do not think users care so much about knowing the exact path the linked section, and perhaps very long links can add extra work for the user to try to parse the gist of what the linked content is about -- this was at least my assumption. 

In this example the full path would actually be `How-to guides > Manage and find data > Use the GUI > Create an account and sign in > Create a NOMAD account`. This seems way to long for me, these intermediate sections/structure are only to help users navigate via the navigation bar, which is irrelevant in this case. 

I would suggest including top-level section + page + section: `How-to guides > ... > Create an account and sign in > Create a NOMAD account` in all cases, with the ... if there are intermediate organizational sections. This way the user knows which top-level section and which page, without getting confused by the subsection of the page naming.

Your point about the search brings a bigger and more impactful issue ... how could we improve the search functionality in the docs??

DanielYang59: thanks sounds good to me and I don't have too strong opinions on this (also I don't have too much knowledge about the designing principles for the docs).

JFRudzinski: would you be able to implement the adjustment globally and update the writing guide?

DanielYang59: emmm I don't think I would have the capacity to read through all docs (I assume by "globally" you mean the entire docs repo?) but I could ask LLM to do this, does this sound good?

Also I assume fixing links (this PR) would be a separate task than changing these pointers, so if LLM is fine I would pick this up once this PR is done

JFRudzinski: Yes, I definitely meant with LLM, thank you!

DanielYang59: Ok I would pick that up once this PR is merged

DanielYang59 (approved):

PR #329, opened 2026-07-27
327 how to write an app refactor

### LINKED ISSUE #327: How to write an app refactor
This page should be refactored in preparation for the new GUI:

- [x] The search quantities no longer need to be explicitly loaded. The docs should reflect on this.
- [x] Add more detailed examples of some core concepts that are easy to copy-paste

lauri-codes:


lauri-codes:
@lukaspie: Maybe you could take a look at this PR to see if something should be changed/added.

docs/howto/plugins/types/apps.md
```diff
@@ -73,54 +78,25 @@ The definition fo the actual app is given as an instance of the `App` class spec
...
+### Search quantities
 
-- Python schemas are identified by the python path for the class that inherits from `Schema`. For example, if you have a python package called `nomad_example`, which has a subpackage called `schema_packages`, containing a module called `mypackage.py`, which contains the class `MySchema`, then the schema name will be `nomad_example.schema_packages.mypackage.MySchema`.
-- YAML schemas are identified by the entry id of the schema file together with the name of the section defined in the YAML schema. For example if you have uploaded a schema YAML file containing a section definition called `MySchema`, and it has been assigned an `entry_id`, the schema name will be `entry_id:<entry_id>.MySchema`.
+Apps can only display and work on values that are indexed in the NOMAD search index. Not all values can/should be meaningfully indexed, as they would otherwise overcrowd and slow down the search functionality. By default, **apps can work on scalar quantities that are present in the schemas**, as they are automatically indexed. This means that e.g. values from lists or multidimensional arrays are not available for search.
```
lukaspie:
```suggestion
Apps can only display and work on values that are indexed in the NOMAD search index. Not all values can/should be meaningfully indexed, as they would otherwise overcrowd and slow down the search functionality. By default, **apps can work on scalar quantities that are present in the schemas**, as they are automatically indexed. This means that, for example, values from lists or multidimensional arrays are not available for search.
```

docs/howto/plugins/types/apps.md
```diff
@@ -73,54 +78,25 @@ The definition fo the actual app is given as an instance of the `App` class spec
...
-The quantities from schemas may be included or excluded by using the [`SearchQuantities`](#app-reference) field in the app config. This option supports a wildcard/glob syntax for including/excluding certain search quantities. For example, to include all search quantities from the Python schema defined in the class `nomad_example.schema_packages.mypackage.MySchema`, you could use:
-
-```python
-search_quantities = SearchQuantities(
-    include=['*#nomad_example.schema_packages.mypackage.MySchema']
-)
-```
-
-The same thing for a YAML schema could be achieved with:
+Quantities that are indexed and can thus be searched are in this document referred to as **search quantities**, and the app configuration often refers to them. E.g. to display a specific search quantity as an app column, one would define it like this:
```
lukaspie: The app column was not really mentioned before (only in the snippet above), so maybe you can link to its explanation later on the page.

docs/howto/plugins/types/apps.md
```diff
@@ -147,31 +123,61 @@ The `menu` field controls the structure of the menu shown on the left side of th
...
-- `MenuItemDefinitions`: Shows a tree of available definitions from which items can be selected for the query.
-- `MenuItemCustomQuantities`: Form for querying custom quantities coming from any schema.
-- `MenuItemNestedObject`: Used to group together menu items so that their query is performed using an Elasticsearch nested query. Note that you cannot yet use nested queries for search quantities originating from custom schemas.
+<!-- `#menu_1` is the generated `#### Menu` heading; `### Menu` above takes `#menu`. -->
+- [`Menu`](#menu_1): Defines a nested submenu.
+- [`MenuItemTerms`](#menuitemterms): Used to display a set of possible text options.
+- [`MenuItemHistogram`](#menuitemhistogram): Histogram of numerical values.
+- [`MenuItemPeriodicTable`](#menuitemperiodictable): Displays a periodic table.
+- [`MenuItemVisibility`](#menuitemvisibility): Controls for the query visibility.
+- [`MenuItemNestedObject`](#menuitemnestedobject): Used to group together menu items so that their query is performed using an Elasticsearch nested query. Note that you cannot yet use nested queries for search quantities originating from custom schemas.
```
lukaspie, line 144: There are many more below like `MenuDefinitions`, `MenuOption`, etc. 

Are these intentionally not mentioned?

docs/howto/plugins/types/apps.md
```diff
@@ -147,31 +123,61 @@ The `menu` field controls the structure of the menu shown on the left side of th
...
 
-### Dashboard
+### Widgets
```
lukaspie, line 146: Any reason for the renaming here? Just to avoid the conflict with the Dashboard EntryPoint? It is a bit weird because the actual app configuration is still called `Dashboard`, but then there are these `Widget*` classes inside. One would possibly expect that the top-level item should be called `Widget`, similarly to how `Menu` has `Menu*` classes inside.

Maybe we could in the future define `Widget = Dashboard` in the code and then use `Widget` publically?

docs/howto/plugins/types/apps.md
```diff
@@ -147,31 +123,61 @@ The `menu` field controls the structure of the menu shown on the left side of th
...
 
- The Dashboard field controls the content of the main search interface window. Different widgets can be added which contain terms or numerical information and can be controlled in size and position. There are 4 different types of Widgets:
+The app may display widgets which contain terms or numerical information and can be controlled in size and position. Here is an example of a widget configuration:
```
lukaspie: terms or numerical information seems to be quite vague

docs/howto/plugins/types/apps.md
```diff
@@ -147,31 +123,61 @@ The `menu` field controls the structure of the menu shown on the left side of th
...
+```python
+--8<-- "examples/plugins/columns.py"
+```
+<!-- fmt: on -->
+
+See the [`Column`](#column) and [`Format`](#format) options for more details.
+
+### Row actions
+
+You may add specific, always-visible actions for each result table row. Here is an example of two actions, one opening up a NORTH tool, and the other for opening up a link.
```
lukaspie: We could explain what we mean by "action" here

lukaspie: We kind of do that already with the example below, but maybe in a more general sense

examples/plugins/columns.py
```diff
@@ -1,11 +1,17 @@
-from nomad.config.models.ui import Column
+from nomad.config.models.ui import Column, Format
```
lukaspie, line 1: Are these intentionally different to the overall app in `app.py`?

examples/plugins/app.py
```diff
@@ -28,22 +27,15 @@
...
         readme='Here is a much longer description of this app.',
-        # If you want to use quantities from a custom schema, you need to load
-        # the search quantities from it first here. Note that you can use a glob
-        # syntax to load the entire package, or just a single schema from a
-        # package.
-        search_quantities=SearchQuantities(
-            include=['*#nomad_example.schema_packages.mypackage.MySchema'],
-        ),
         # Controls which columns are shown in the results table
         columns=[
```
lukaspie, line 32: The first thing we define here is the columns, whereas in the docs page we go Menu -> Widgets -> Colums. Maybe we could describe them in the order as they show up in the app.

lukaspie (approved):
Overall very good and much appreciated. I had some minor comments. I will approve already, you can address the comments as you see fit,

docs/howto/plugins/types/apps.md
lauri-codes: This is intentional: I don't mention the deprecated ones here.

docs/howto/plugins/types/apps.md
lauri-codes: Yes the renaming is due to the clash with the new Dashboard plugin entry point type.

We should migrate the app configuration to use another term, but not sure yet what we call it.

docs/howto/plugins/types/apps.md
lauri-codes: I now improved this a bit.

examples/plugins/columns.py
lauri-codes: Now unified.

examples/plugins/app.py
lauri-codes: Yes

docs/howto/plugins/types/apps.md
lauri-codes: Agreed, now fixed.

docs/howto/plugins/types/apps.md
lukaspie: I would think `Widget` is just fine, that's what such small, self-contained pieces of code that serve a specific content are often called. But I let you make that decision.

PR #338, opened 2026-08-11
Add CNAME

DanielYang59:
https://github.com/orgs/community/discussions/159544#discussioncomment-13179908

### DIFF
--- a/docs/CNAME
+++ b/docs/CNAME
@@ -1,0 +1,1 @@
+docs.nomad-lab.eu


blueraft:
This doesn't add them to the docs that are built. 


It probably needs to be something like this:

```yaml
      - name: Build MkDocs site
        run: |
          uv run \
            --with 'nomad-lab[dev,infrastructure] @ https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR.git' \
            mkdocs build

      - name: Configure custom domain
        run: |
          echo "docs.nomad-lab.eu" > site/CNAME
```

DanielYang59:
I just tried it seems to add it (locally)... a bit unsure about github actions

```
>>> uv run mkdocs build --strict

>>> cat site/CNAME
docs.nomad-lab.eu
```

Just realized [`mkdocs` documented this behaviour and fix](https://www.mkdocs.org/user-guide/deploying-your-docs/#custom-domains):



> GitHub Pages includes support for using a [Custom Domain](https://help.github.com/articles/adding-or-removing-a-custom-domain-for-your-github-pages-site) for your site. In addition to the steps documented by GitHub, you need to take one additional step so that MkDocs will work with your custom domain. You need to add a CNAME file to the root of your [docs_dir](https://www.mkdocs.org/user-guide/configuration/#docs_dir). The file must contain a single bare domain or subdomain on a single line (see MkDocs' own [CNAME file](https://github.com/mkdocs/mkdocs/blob/master/docs/CNAME) as an example). 

> You may create the file manually, or use GitHub's web interface to set up the custom domain (under Settings / Custom Domain). If you use the web interface, GitHub will create the CNAME file for you and save it to the root of your "pages" branch. So that the file does not get removed the next time you deploy, **you need to copy the file to your docs_dir**. With the file properly included in your docs_dir, MkDocs will include the file in your built site and push it to your "pages" branch each time you run the gh-deploy command.

blueraft:
Well let's try it out

